PR #66, opened 2022-07-14
Update dbt_cloud_connection.md

felipefrancisco:
Update the connection docs so it reads "connection" instead of "environment" on the `project_id` bullet point


felipefrancisco:
@GtheSheep thank you for the feedback, updated 👍🏽

GtheSheep:
@felipefrancisco thanks for this! Spotting my copy + paste 😹

GtheSheep (approved):

PR #84, opened 2022-09-28
Add database, role, warehouse to Snowflake credentials resource

mwufigma:
Updated tests + `make test`

Also tested via Postman to make sure the API accepts `role`, `database`, `warehouse` with my own DBT Cloud account.


pkg/resources/snowflake_credential.go
```diff
@@ -102,14 +102,17 @@ func resourceSnowflakeCredentialCreate(ctx context.Context, d *schema.ResourceDa
 	isActive := d.Get("is_active").(bool)
 	projectId := d.Get("project_id").(int)
 	authType := d.Get("auth_type").(string)
+	database := d.Get("database").(string)
```
GtheSheep, line 105: these need to be added to the resource schema further up in this file too (causing tests to fail) 😄

GtheSheep (approved):

PR #86, opened 2022-09-29
Add self deferring functionality on dbt_cloud_job

mwufigma:
The logic is a little more complicated than other configurations. Summarized below:
```
Create Job:
if self_deferring:
  create a POST request to create a job, then send a subsequent POST request to update it with it's own ID

Update Job:
if self_deferring changed:
  if self_deferring == true:
    set deferring_job_id to itself
  else:
    set deferring_job_id to the actual deferring_job_id value in the resource

Reading from remote resource:
if deferring_job_id == job.ID:
  set self_deferring in ResourceSchema to true
```

Tested with `make test` (Though I noticed there were no unit tests for Jobs, only acceptance tests, which I will transition over to you)


pkg/resources/job.go
```diff
@@ -133,6 +133,11 @@ var jobSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Job identifier that this job defers to",
 	},
+	"self_deferring": &schema.Schema{
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Description: "Whether this job defers on a previous run of itself",
```
GtheSheep, line 139: I think there's an arg to the effect of `ConflictsWith`, could you add that to conflict with the deferring_job_id plz?

pkg/resources/job.go
```diff
@@ -284,7 +293,7 @@ func resourceJobUpdate(ctx context.Context, d *schema.ResourceData, m interface{
 		d.HasChange("target_name") || d.HasChange("execute_steps") || d.HasChange("run_generate_sources") ||
 		d.HasChange("generate_docs") || d.HasChange("triggers") || d.HasChange("schedule_type") ||
 		d.HasChange("schedule_interval") || d.HasChange("schedule_hours") || d.HasChange("schedule_days") ||
-		d.HasChange("schedule_cron") || d.HasChange("deferringJobId") {
+		d.HasChange("schedule_cron") || d.HasChange("deferringJobId") || d.HasChange("selfDeferring") {
```
GtheSheep, line 296: should be the self_deferring I think? (the deferringJobId prior is wrong 😬 )

mwufigma: Fixed both!

pkg/data_sources/job.go
```diff
@@ -36,6 +36,12 @@ var jobSchema = map[string]*schema.Schema{
 		Computed:    true,
 		Description: "ID of the job this job defers to",
 	},
+	"self_deferring": &schema.Schema{
+		Type:        schema.TypeBool,
+		Computed:    true,
+		Default:     false,
```
GtheSheep, line 42: tests are complaining that this can't default as it's computed anyway in the data source 😅

pkg/resources/job.go
```diff
@@ -360,6 +371,16 @@ func resourceJobUpdate(ctx context.Context, d *schema.ResourceData, m interface{
 			deferringJobId := d.Get("deferring_job_id").(int)
 			job.Deferring_Job_Id = &deferringJobId
 		}
+		// If self_deferring has been toggled to true, set deferring_job_id as own ID
+		// Otherwise, set it back to what deferring_job_id specifies it to be
+		if d.HasChange("self_deferring") {
+			if d.Get("self_deferring") == true {
+				job.Deferring_Job_Id = job.ID
```
GtheSheep, line 378: maybe set this to `jobId` instead? the [tests](https://github.com/GtheSheep/terraform-provider-dbt-cloud/actions/runs/3178279205/jobs/5179609407) don't like some kind of ID

mwufigma: I don't think the acceptance tests actually cover `resourceJobUpdate` do they? I might be mistaken. Based on the error output I see in the tests, it seems like only the following methods were called:
```
resourceJobCreate -> Client.CreateJob -> Client.UpdateJob
```
These lines: https://github.com/GtheSheep/terraform-provider-dbt-cloud/pull/86/files/2f9b684d9d1af9832509097c8a6a49c45ac14f3a#diff-35de52af826f07e23da639c22db938021a3e90c0eaa0b4cf27fc4f83c2d295a0R182-R184
seem to be the ones in this call stack that have to do with passing in a job's own ID into an update. Not sure what the issue is, but I'll get a dbt trial account and test there.

mwufigma: I'll also add an additional acceptance test step to test resource update

GtheSheep: from [here](https://github.com/GtheSheep/terraform-provider-dbt-cloud/blob/36361f8381fa5d6bb8f9d732984dba18455eb7ac/pkg/resources/job_acceptance_test.go#L34) down is essentially testing updating a job resource, just doesn't check the values actually which it probably should.

But yeah, it's those update lines in the create that it's not liking, probs worth printing out the ID it's trying to update and the IDs used in that block as it's not matching somewhere 😬 seems like a minor issue tho tbf

PR #100, opened 2022-11-30
Feature - Add support for dbt cloud artefacts (source freshness and docs)

### LINKED ISSUE #98: Add docs/ freshness job params to projects

b-per:
Closes #98 

This is my attempt at adding links to source freshness and docs jobs (which I called artefacts) at the project level.
I also updated the data source to contain the job ids for docs and source freshness.

Behind the scenes, the implementation is quite close to the `project_connection` one, calling the `xxxProject()` methods.

One thing that I wasn't sure how to manage properly was to get the ids set to `nil` instead of `0` when there is no `job_id` defined.

This is still WIP but I'd be keen for any feedback until now.

Still to be done:
- [x] add tests (and figure out how those work in Go/Terraform)
- [x] update docs


pkg/resources/project_artefacts.go
```diff
@@ -0,0 +1,180 @@
...
+	project, err := c.GetProject(projectIDString)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if docsJobID != 0 {
+		project.DocsJobId = &docsJobID
+	} else {
+		project.DocsJobId = nil
+	}
```
b-per, line 70: Currently handling `nil` that get automatically changed to `0` that way but it doesn't feel right...

pkg/resources/project_artefacts.go
```diff
@@ -0,0 +1,180 @@
...
+			return diag.FromErr(err)
+		}
+
+		if d.HasChange("docs_job_id") {
+			docsJobId := d.Get("docs_job_id").(int)
+			if docsJobId != 0 {
+				project.DocsJobId = &docsJobId
+			} else {
+				project.DocsJobId = nil
+			}
```
b-per, line 137: Same on the `Update` side

b-per:
Added some acceptance tests (they ran OK for me) and updated the docs
image

I am not a huge fan of the fact that if a freshness or docs job is not setup it shows with the `id` 0 in the state (when it should actually be null) but otherwise it all seems to be working fine!

b-per:
Hi @GtheSheep ! I am not sure if you saw this PR so this is just a quick ping 😃

pkg/resources/project_artefacts.go
GtheSheep: Think you may be able to deal with this by having them as optional/ pointers on the Project struct? then using omitempty in the json options?

pkg/dbt_cloud/project.go
```diff
@@ -16,6 +16,8 @@ type Project struct {
 	RepositoryID           *int    `json:"repository_id,integer,omitempty"`
 	State                  int     `json:"state"`
 	AccountID              int     `json:"account_id"`
+	FreshnessJobId         *int    `json:"freshness_job_id"`
```
GtheSheep, line 19: can these be sent to the create/ update endpoints? i.e. are they just extra fields for the project endpoint or is it necessary to manage as a project_artefacts resource at all?

GtheSheep:
hey @b-per! sorry for the slow response, have chucked a couple of comments on here now when you get a chance, thanks again 😸

pkg/resources/project_artefacts.go
b-per: I have tried it but it didn't work. After some (way too much) digging, it looks like it is a limitation of of [SDKv2](https://stackoverflow.com/questions/73341494/is-it-possible-to-differentiate-between-a-zero-value-and-a-removed-property-in-t) that has been "fixed" in the new (but not stable yet) [Terraform Plugin Framewrok](https://developer.hashicorp.com/terraform/plugin/framework).

So despite the fact that it is not super clean I don't think that we can get around the small logic.

b-per:
Thanks for the review 👍 
I have added additional context to your comments!

b-per:
@GtheSheep , no rush, but I'd need you to approve the workflow for this.
Let me know if you have any other feedback/question from the ones I posted last time.

GtheSheep:
@b-per - sorry, forgot about this! will take a look this week and test/ merge

pkg/resources/project_artefacts.go
GtheSheep: @b-per - I've done the migration for the Plugin Framework on another provider, so will do it on this one at a later date and we can deal with it then

pkg/dbt_cloud/project.go
GtheSheep: Ahh I see, the same reason resources [like this](https://registry.terraform.io/providers/GtheSheep/dbt-cloud/latest/docs/resources/dbt_cloud_project_connection) exist... ok, makes sense, let's do it like this for now so that it's at least possible 😸

pkg/resources/project_artefacts_acceptance_test.go
```diff
@@ -0,0 +1,171 @@
...
+func testAccDbtCloudProjectArtefactsResourceBasicConfig(projectName, environmentName, jobName string) string {
+	return fmt.Sprintf(`
+resource "dbt_cloud_project" "test_artefacts_project" {
+	name = "%s"
+}
+
+resource "dbt_cloud_environment" "test_job_environment" {
+	project_id = dbt_cloud_project.test_artefacts_project.id
+	name = "%s"
+	dbt_version = "0.21.0"
```
GtheSheep, line 59: @b-per - 1.0.0 ? 😅 otherwise this LGTM 👌

Issue #108, opened 2023-02-15
Error when creating new environment variables using the dbt_cloud_environment_variable resource


subrasub:
I am running into an issue while attempting to create an environment variable using the `dbt_cloud_environment_variable` resource. 

Running `terraform plan` was successful, however, I am unable to apply these changes, as the endpoint is returning a 500 error. Upon investigating the code for this resource, I realized it was using dbt-cloud v3 API, so I attempted to manually create a POST request to the endpoint with the help of v3 API docs, but I was unable to find it. https://github.com/GtheSheep/terraform-provider-dbt-cloud/blob/e65faea6fa1226a129e208d2f33a76582611a94d/pkg/dbt_cloud/environment_variable.go#L87

I am not sure if the endpoint has been deprecated or if there is another issue at hand. Also, I am curious as to why our CircleCI build pipeline did not flag this error - `terraform apply` command was successful, but had a 500 error within. I would appreciate any help!

GtheSheep:
Hey @subrasub - Could you please provide a little more context around how this failed? As far as I can tell, as long as your project names match case sensitively then there's no issue? You could try making a request like:
```
import json, requests
response = requests.post(
    'https://cloud.getdbt.com/api/v3/accounts/<account ID>/projects/<project ID>/environment-variables/bulk/',
    headers={
      'Accept': 'application/json',
      'Content-Type': 'application/json',
      'Authorization': 'Token <token>'
  },
    json={
        'env_var': {
            'new_name': 'DBT_<env var name>',
            'project': '<project level var>',
            '< env 1 >: '< val for env 1 >',
            '< env 2 >: '< val for env 2 >',
            ...
        }
    }
)
response.json()
```

Could you also please provide any config terraform (with values redacted) where possible so I can attempt to reproduce, please?

PR #109, opened 2023-02-16
Update dbt_cloud_environment_variable.md

subrasub:
Add a note to title-case environment names. If we don't title-casing the environment names, it can throw a 500 error when we run `terraform apply`. For eg: if we are setting the environment variables for the Production env, the environment name should also be `Production`.

Related - https://github.com/GtheSheep/terraform-provider-dbt-cloud/issues/108#issue-1586440179

### DIFF
--- a/docs/resources/dbt_cloud_environment_variable.md
+++ b/docs/resources/dbt_cloud_environment_variable.md
@@ -18,7 +18,9 @@ the same time as the environment variables, it's recommended to use the `depends
 
 ### Required
 
-- `environment_values` (Map of String) Map from environment names to respective variable value, a special key `project` should be set for the project default variable value
+- `environment_values` (Map of String) Map from environment names to respective variable value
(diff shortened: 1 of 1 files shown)


GtheSheep:
Hey @subrasub - thanks for this! But these descriptions are set in the [resource object](https://github.com/GtheSheep/terraform-provider-dbt-cloud/blob/82ddf802fb9da14ed6b848ace33982a396e95fe5/pkg/resources/environment_variable.go#L42), the markdown is then generated from them using `make doc`, could you update? thanks!

GtheSheep:
Also, reading your comment, I'm not sure this is clear? Are you just saying the environment names are case sensitive and need to exactly match how the environment is configured?
i.e. Environment name `Production` can't have env vars under `production`?

GtheSheep:
Just tested this out and it works for non title-cased envs too, i.e. `staging`

PR #119, opened 2023-03-15
Feature/allow native GitHub repos

### LINKED ISSUE #105: git_clone_strategy = "github_app" not supported
Hi,
we are running a dbtCloud - Snowflake - GitHub - Terraform setup, and are really happy about this provider. Thanks a lot for all the work that has been put into it!
We are experiencing trouble when connecting to our GitHub repo, as the `git_clone_strategy = "github_app"` for some reason doesn't work. Instead, the repo gets connected with `git_clone_strategy = "deploy_key"`. Unfortunately, that means that we can't set CICD up for any of the jobs. However, if we connect the GitHub repo directly in dbtCloud, we can without any trouble setup a job with`"github_webhook" : true` via terraform.
not sure if this is a bug that can easily be solved, but would be great to get rid of that manual step in between!
I'm happy to provide more details, if needed.

PS: Same seems to hold true for the option of custom branches in resource dbt_cloud_environment: `use_custom_branch = true` doesn't work until set manually in dbtCloud.

b-per:
Fixes #105 

This PR adds the ability to connect to connect a repo using the native GitHub integration rather than the `deploy_token` one. This is similar to the approach already created for GitLab.

I have added a new test (that would work just for me though as the `github_installation_id` is currently hard code. The 2 tests are passing:
```
❯ /opt/homebrew/bin/go test -timeout 200s -run ^TestAccDbtCloudRepositoryResource$ github.com/gthesheep/terraform-provider-dbt-cloud/pkg/resources -v
=== RUN   TestAccDbtCloudRepositoryResource
--- PASS: TestAccDbtCloudRepositoryResource (37.85s)
PASS
ok      github.com/gthesheep/terraform-provider-dbt-cloud/pkg/resources 38.443s
(base) 
```

I have also added examples on how to create a `dbt_cloud_repository` resource and how someone could find their `github_installation_id`.

A key change I did as well was to move most of the repository fields to `ForceNew: true,` as I am not sure that the API lets us modify an existing repository. From the UI point of view, changing a repository requires in most of the cases disconnecting the existing repo and connecting a new one, similar to the `ForceNew` approach.


examples/resources/dbt_cloud_repository/resource.tf
```diff
@@ -0,0 +1,29 @@
...
+  remote_url = "<gitlab-group>/<gitlab-project>"
+  gitlab_project_id = 8765
+}
+
+
+resource "dbt_cloud_repository" "deploy_repo" {
+  project_id = 3456
+  remote_url = "git://github.com/<github_org>/<github_repo>.git"
+  git_clone_strategy = "deploy_token"
+}
```
GtheSheep, line 29: new line at EOF plz 😅

examples/resources/dbt_cloud_repository/resource.tf
```diff
@@ -0,0 +1,29 @@
...
+# dbt_cloud_api_result = await (fetch('https://<dbt_cloud_url>/api/v2/integrations/github/installations/').then(res => res.json()));
+# console.log("github_application_id: " + dbt_cloud_api_result.filter(res => res["access_tokens_url"].includes("github"))[0]["id"]);
+# ```
+
+# Alternatively, you can go to the page https://<dbt_cloud_url>/api/v2/integrations/github/installations/ and read the value of "id"
+
+resource "dbt_cloud_repository" "github_repo" {
+  project_id = 1234
+  remote_url = "git://github.com/<github_org>/<github_repo>.git"
+  github_installation_id = 9876
```
GtheSheep, line 13: just so I understand fully, this is created in dbt Cloud when adding a Github repo? Does it not come in the response for the POST request for repositories?

b-per: When using the "GitHub" choice when liking a repo, instead of using the "Git Clone" approach, dbt Cloud connects to the repo via a GH App.
image

When using the UI, `github_installation_id` is automatically added to the first `POST` request. It is also returned if we `GET` the repo.

examples/resources/dbt_cloud_repository/resource.tf
b-per: Sure!

b-per:
HI @GtheSheep !

Is there a way for me to help debugging the failing unit test? Clicking on "Details" doesn't show any detail for me.

GtheSheep:
Hey @b-per! I think you can navigate to the run via the Actions part of the repo, lmk if not and I'll try and dig it out, not sure why it doesn't come through the details bit!

b-per:
I can see difference tests on the Action tab but none fort this specific PR.

I just see this
image

Drilling into
image

And then
image

b-per:
For other actions, like [this one](https://github.com/GtheSheep/terraform-provider-dbt-cloud/actions/runs/4451424935), I can see the details.

(BTW those other ones are failing because dbt Cloud deprecated dbt-Core < 1.0 versions)

examples/resources/dbt_cloud_repository/resource.tf
GtheSheep: So we could, in theory, retrieve the ID from `https://<dbt_cloud_url>/api/v2/integrations/github/installations/` in the provider if we wanted to? Rather than the user having to go and get it after doing a manual connection? Or am I misunderstanding the flow and that this already expects dbt Cloud <-> Github to be linked in order for this ID to exist?

b-per:
Hi @GtheSheep !
I have pulled from `main` to retrigger the tests, and I can see the results now!

I think that there is a small issue with the acceptance tests, the env var `DBT_CLOUD_ACCOUNT_ID` doesn't seem to be set correctly.

GtheSheep:
@b-per - Ahhh damn, this has made me realise why I had all of the ok-to-test stuff in place, forks don't have access to secrets in PRs 😬 I've changed to trigger on `pull_request_target` to try this out where it will require approval to run also.

In the meantime I think the only question is around those `github_installation_id`s?

b-per:
Sure, what is your question about `github_installation_id`?

I have added instructions on how to retrieve it in the example for the `dbt_cloud_repository` but I don't know if there is anything else to add.

The `github_installation_id` will be unique for a given account, so it will need to be retrieved only once.

GtheSheep:
Was just wondering if it was worth adding a call to retrieve it for the user, but could always do that later, it's not a sensitive piece of info, right?

PR #131, opened 2023-04-03
Create postgres redshift credential resource

dave-connors-3:
This PR adds resources for Postgres Credential objects, which supports both postgres and Redshift connections in dbt cloud!


pkg/resources/postgres_credential.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := d.Set("project_id", postgresCredential.Project_Id); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("default_schema", postgresCredential.Default_Schema); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("username", postgresCredential.Username); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("password", postgresCredential.Password); err != nil {
```
GtheSheep, line 150: Yeah so assuming `c. GetPostgresCredential` does not return the password from the API, this will always be set to "" which is what's causing the resource test to fail, if you want to set it during create (and we should add a warning about storing passwords in state files as it seems fairly standard to do so, as a log or in the docs), you'll probably need to access it from the resourceData with `d.get("password").(string)` or something instead?

pkg/data_sources/postgres_credential_acceptance_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDbtCloudPostgresCredentialDataSource(t *testing.T) {
+
+	randomProjectName := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+
+	config := postgres_credential(randomProjectName, "moo", "baa", "maa", 64)
```
GtheSheep, line 15: 😹 "moo" "baa" "maa"

b-per:
@dave-connors-3 : [this branch](https://github.com/b-per/terraform-provider-dbt-cloud/tree/update-postgres-code) in my fork is passing the tests, you can just replicate my last commits.

Also, pull from `main` as well before pushing back your branch as I think the requests have a longer delay before failing now. (this is for the failing test unrelated to your change)

GtheSheep (changes requested):
Looks pretty good to me! Could you run 'make doc' and add the docs for this new data source/ resource too please 🙂

pkg/resources/postgres_credential.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/gthesheep/terraform-provider-dbt-cloud/pkg/dbt_cloud"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var (
+	types = []string{
+		"postgres",
+		"redshift",
```
GtheSheep, line 18: Oh is it identical to Postgres? Well that makes things much easier 😅

GtheSheep (approved):

PR #133, opened 2023-04-25
Update docs

b-per:
This PR adds small update to the docs but has a big `diff` due to many files getting renamed in the process.

This
- fixes some of the examples that were not showing (fixed by providing `generate --provider-name dbt` to `tfplugindocs`)
- updated a few other examples that were not working or just missing
- changed `dbt` to lowercase everywhere it was still `DBT` 😄
- moved the example for postgres_credential to the example folder 
- mentions the fact we can use env vars on the main page


docs/resources/cloud_postgres_credential.md
```diff
@@ -25,15 +24,14 @@ resource "dbt_cloud_postgres_credential" "test_credential" {
...
-
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
 ### Required
 
 - `default_schema` (String) Default schema name
 - `project_id` (Number) Project ID to create the Postgres/Redshift/AlloyDB credential in
-- `type` (String) Type of connection. One of (postgres/redshift) use postgres for alloydb connections
+- `type` (String) Type of connection. One of (postgres/redshift)
```
GtheSheep, line 34: was this not true?

examples/resources/dbt_cloud_postgres_credential/resource.tf
```diff
@@ -0,0 +1,9 @@
+resource "dbt_cloud_postgres_credential" "test_credential" {
+    is_active = true
+    project_id = dbt_cloud_project.test_project.id
+ type = "postgres"
+    default_schema = "%s"
+    username = "%s"
+    password = "%s"
+    num_threads = 3
+}
```
GtheSheep, line 9: nit: new line at EOF

docs/index.md
```diff
@@ -1,12 +1,12 @@
...
 # generated by https://github.com/hashicorp/terraform-plugin-docs
-page_title: "dbt-cloud Provider"
+page_title: "dbt Provider"
 subcategory: ""
 description: |-
   
 ---
 
-# dbt-cloud Provider
+# dbt Provider
```
GtheSheep, line 9: I guess this is the downside, if it's now dbt provider, but not the end of the world

b-per: Yes, I think that getting it to be dbt-cloud but still work with templating might require renaming a few things.

GtheSheep: Yeah, I think that's something I was partly trying to solve by moving the the new framework where the name is in a metadata object

GtheSheep (approved):
The version file doesn't do anything right now, had the intention to use it in releasing but never mind 😅 in case you're concerned about bumping it

b-per:
### Comment on the version number

Thanks for the update on the version number. Right now in the end it is still useful when doing a `make install` and push changes to a specific version but I will keep in mind that it is not used otherwise.

### Comment on this PR 

About this PR, I have second thoughts about using just `dbt` when generating the docs. I was eavesdropping other providers and most (if not all of them) use single words. The ones being "multiple words" just concatenate them into one. For example [ionoscloud](https://registry.terraform.io/providers/ionos-cloud/ionoscloud/latest) or [octopusdeploy](https://registry.terraform.io/providers/OctopusDeployLabs/octopusdeploy/latest)

So, I think that the docs could/should be generated with `generate --provider-name dbtcloud` (instead of the current PR changing it to `generate --provider-name dbt`)

PR #139, opened 2023-05-17
Feature - Add support for outbound webhooks

### LINKED ISSUE #137: Feature - Support creation/deletion of Webhooks
It is now possible to create [outbound webhooks](https://docs.getdbt.com/docs/deploy/webhooks) in dbt Cloud.
The provider could be used to create those.

I am working on it and will have a PR in the next few days.

b-per:
Fixes #137

Will most likely conflict on `provider.go` with the PR for the service token.
We will also need to know if we want to merge #133 before or after as this will require som slight docs rework

### datasource acceptance test passing
SCR-20230517-qikz

### resource acceptance test passing
SCR-20230517-rxde


b-per:
The CI issues seems to be about a problem connecting to dbt Cloud and not an issue with the new tests.

pkg/dbt_cloud/webhook.go
```diff
@@ -0,0 +1,140 @@
...
+	name string,
+	description string,
+	clientUrl string,
+	eventTypes []string,
+	jobIds []int,
+	active bool,
+) (*WebhookRead, error) {
+
+	newWebhook := WebhookWrite{
+		WebhookId:   webhookId,
```
GtheSheep, line 70: shouldn't this not exist until after creation?

pkg/resources/webhook.go
```diff
@@ -0,0 +1,281 @@
...
+		if d.HasChange("job_ids") {
+			jobIDs := make([]int, len(d.Get("job_ids").([]interface{})))
+			for i, step := range d.Get("job_ids").([]interface{}) {
+				jobIDs[i] = step.(int)
+			}
+			webhook.JobIds = jobIDs
+		}
+		if d.HasChange("active") {
+			active := d.Get("active").(bool)
+			webhook.Active = active
```
GtheSheep, line 257: just to check, this doesn't delete it if it's set to not active like some of the other endpoints, right?

pkg/resources/webhook.go
```diff
@@ -0,0 +1,281 @@
...
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     true,
+		Description: "Webhooks active flag",
+	},
+	"hmac_secret": &schema.Schema{
+		Type:        schema.TypeString,
+		Computed:    true,
+		Sensitive:   true,
+		Description: "Webhooks active flag",
```
GtheSheep, line 70: wrong desc I think?

pkg/resources/webhook.go
```diff
@@ -0,0 +1,281 @@
...
+	},
+	"http_status_code": &schema.Schema{
+		Type:        schema.TypeString,
+		Computed:    true,
+		Description: "Webhooks HTTP Status Code",
+	},
+	"account_identifier": &schema.Schema{
```
GtheSheep, line 77: is this the dbt Cloud account ID? Think it exists on the Client which could save having it here?

examples/resources/dbt_cloud_webhook/resource.tf
```diff
@@ -0,0 +1,13 @@
...
+  client_url = "http://localhost/nothing"
+  event_types = [
+    "job.run.started",
+    "job.run.completed"
+  ]
+  job_ids = [
+    1234,
+    5678
+  ]
+}
```
GtheSheep, line 13: nit: New line at EOF

pkg/resources/webhook.go
b-per: Indeed, some copy/paste issue. I will update it.

pkg/dbt_cloud/webhook.go
b-per: It is actually required but with an empty value for the creation of the webhook.

It is set up empty [here](https://github.com/b-per/terraform-provider-dbt-cloud/blob/2e132ec6690c6b8b3f57a4add47de3e98dff52be/pkg/resources/webhook.go#L155)

pkg/resources/webhook.go
b-per: The API doesn't actually allow setting active to `false` and we need to `DELETE` it instead. I remove this from triggering a change to the resource.

GtheSheep (approved):

Issue #146, opened 2023-06-08
Feature - Add support to allow warehouse connections to use PrivateLink


b-per:
Since recently, it is possible to connect dbt Cloud to the Warehouse via PrivateLink for most warehouses (all except BigQuery). 

Creating those via the API require:
1. getting the ID of the PrivateLink endpoint to use (I have some code ready for a datasource for this)
2. modifying the current create/read/update of the resources for connections, for all the warehouse supporting PrivateLink 

Here is an example of payload to create a Snowflake connection using PrivateLink:

```json
{
  "id": null,
  "type": "snowflake",
  "account_id": "<your_account_id>",
  "project_id": "<your_project_id>",
  "name": "<your_connection_name>",
  "state": 1,
  "private_link_endpoint_id": "<your_privatelink_endpoint_id>", <-- this is the new part for point 2.
  "details": {
    "account": "<your_account>.privatelink.snowflakecomputing.com", <-- this is the account with 
    "role": null,
    "database": "db",
    "warehouse": "wh",
    "client_session_keep_alive": false,
    "allow_sso": false,
    "oauth_client_id": null,
    "oauth_client_secret": null
  }
}
```

I believe that for all adapters, the only change is setting a value for `private_link_endpoint_id` and potentially updating other existing fields (like `details.account` here).

PR #147, opened 2023-06-09
Add datasource for privatelink endpoints

b-per:
Fixes part 1 of #146 by allowing people to retrieve information about their PrivateLink endpoints.

Especially, the `id` will be needed for part 2  of the issue, to create DW connections leveraging a PL connection.

As those connections can be read but not created, the acceptance test is behind an env var `DBT_ACCEPTANCE_TEST_PRIVATE_LINK` as it requires reading from existing resources rather than creating new ones.

Locally, with the env var added, the tests pass:
image


pkg/data_sources/privatelink_endpoint.go
```diff
@@ -0,0 +1,88 @@
...
+		Optional:    true,
+		Default:     "",
+		Description: "Given descriptive name for the PrivateLink Endpoint (name and/or private_link_endpoint_url need to be provided to return data for the datasource)",
+	},
+	"type": &schema.Schema{
+		Type:        schema.TypeString,
+		Computed:    true,
+		Description: "Type of the PrivateLink Endpoint",
+	},
+	"private_link_endpoint_url": &schema.Schema{
```
GtheSheep, line 28: Is this not required?

examples/data-sources/dbt_cloud_privatelink_endpoint/data-source.tf
```diff
@@ -0,0 +1,13 @@
+data "dbt_cloud_privatelink_endpoint" "test_with_name" {
+  name = "My Endpoint Name"
+}
+
+data "dbt_cloud_privatelink_endpoint" "test_with_url" {
+  private_link_endpoint_url = "abc.privatelink.def.com"
+
```
GtheSheep, line 7: Nit: blank line 😅

pkg/data_sources/privatelink_endpoint.go
GtheSheep: Ahh I see it needs name or url? Is there a way to validate one or the other is set via the plugin framework?

pkg/dbt_cloud/privatelink_endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+type PrivatelinkEndpointResponse struct {
+	Data   PrivatelinkEndpoint `json:"data"`
+	Status ResponseStatus      `json:"status"`
+}
+
+func (c *Client) GetPrivatelinkEndpoint(endpointName string, privatelinkEndpointURL string) (*PrivatelinkEndpoint, error) {
+
+	if endpointName == "" && privatelinkEndpointURL == "" {
```
GtheSheep, line 31: Ah I see you're enforcing that one has to be provided here

pkg/data_sources/privatelink_endpoint.go
b-per: Yes, it is name or URL, or both.

In the Framework we could use [this](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/resourcevalidator#AtLeastOneOf) to validate but not on the SDK v2. As you saw, I am doing the check when calling the API to make sure that one of them is set.

examples/data-sources/dbt_cloud_privatelink_endpoint/data-source.tf
b-per: I have an empty line at the end locally, not sure why it didn't get pushed

PR #148, opened 2023-06-12
Feature/pl warehouse connection

### LINKED ISSUE #146: Feature - Add support to allow warehouse connections to use PrivateLink

b-per:
Built on top of #147 (we could just merge this one which fixes the merge conflict rather than merging #147 first) which added a datasource for the PL connections.

Fixes #146 :Allows use to use `private_link_endpoint_id` to set up PrivateLink connections to their Data Warehouses.

As PL connections can't be created programatically the test was running successfully locally using an existing PL connection but it requires the env var `DBT_ACCEPTANCE_TEST_PRIVATE_LINK` that should not be set on Acceptance tests in CI.


pkg/resources/connection.go
```diff
@@ -56,6 +56,13 @@ func ResourceConnection() *schema.Resource {
 				Required:     true,
 				Description:  "The type of connection",
 				ValidateFunc: validation.StringInSlice(connectionTypes, false),
+				ForceNew:     true,
```
GtheSheep, line 59: Just double checking this is intentionally on type rather than private link? (Just as that's what this PR relates to)

GtheSheep (approved):

PR #159, opened 2023-06-23
Add guide

b-per:
The guide got removed when generating docs. Adding it again.


b-per:
Cancelling CI as there is no logic change, just docs

PR #168, opened 2023-08-02
✨ Add support for notifications (int/ext)

b-per:
Add the ability to define email notifications in dbt Cloud both for internal dbt Cloud users and external email addresses.

This PR allows what is defined in [this post](https://discourse.getdbt.com/t/how-to-configure-external-user-email-notifications-in-dbt-cloud/8393) to be defined in Terraform.


b-per:
Cancelled the workflow as I just updated 1 doc page and it passed in the step before

PR #197, opened 2023-09-20
fix: Update GetGroup function to get by id rather than list and find

SudoSpartanDan:
We found that once you get over 100 groups in DBT, the GetGroup function was failing to find some group. This is because the function uses the list endpoint, which is paged and defaults to 100 groups as the limit to return.

This PR adjusts the function to use the Retrieve Group endpoint: https://docs.getdbt.com/dbt-cloud/api-v3#/operations/Retrieve%20Group

For some reason, the API docs say data returns as a list but actually testing it, it came back as a JSON. #shrug, I think it's probably a doc that just needs updating.

### DIFF
--- a/pkg/dbt_cloud/group.go
+++ b/pkg/dbt_cloud/group.go
@@ -45,7 +45,7 @@ type GroupPermissionListResponse struct {
 }
 
 func (c *Client) GetGroup(groupID int) (*Group, error) {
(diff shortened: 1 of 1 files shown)


b-per:
Thanks for this!

There are actually different parts in the provider where things might break when we reach more than 100 items. I will create a new issue to track those and either try to switch to an endpoint that queries one specific ID or loop through the results pages.

I will release a new version tomorrow I think with this and maybe a couple of extra changes.

PR #206, opened 2023-10-19
Fixes-example-problems

will-sargent-dbtlabs:
This PR refines some of the examples provided, to be more consistent in their resource naming.

It also allows for cleaner  cut->paste into a single new main.tf script, to just generally work.

I also added an additional note at the top, where relevant, that further clarifies the "legacy provider" comment, as I found it a bit confusing at first, trying to understand if I needed to do something or not (not a legacy customer).

There is probably some additional work still to be done here, especially on the `import.sh` front, making sure that aligns.


b-per:
I think that this is a good step forward and I am happy to merge it as-is.
We can revisit/update later if need be.

The only thing I did in addition was to do a `make doc` to regenerate the `docs` folder based on the examples provided.

Issue #221, opened 2023-11-16
Removing an `environment_values` for `dbtcloud_environment_variable` doesn't remove it in dbt Cloud


b-per:
## Describe the bug
Removing an `environment_values` for an existing `dbtcloud_environment_variable` resource doesn't remove it in dbt Cloud

## Error message
No error message, but the env variable is not removed

## Expected behavior
The env var is removed

b-per:
Setting the value to `""` doesn't do anything either

b-per:
The current resource uses an "older" API technique where we do a `PUT` on `https://emea.dbt.com/api/v3/accounts/<account>/projects/<project>/environment-variables/bulk/` with the body

```
{
    "env_vars": {
        "DEV": "2",
        "PROD": "",
        "name": "DBT_TF_ENV_VAR",
        "project": "1"
    }
}
```

If some keys are removed, or provided empty inputs, they are not removed in dbt Cloud.

Some options to fix it would be:
1. try to add some logic where if a key is removed from `env_vars` we force the destroy and recreation of the resource
2. and/or creating a new resource where the resource grain would be at the `env_var` + scope (e.g. project or env name) rather than just `env_var`, allowing to destroy specific `env_var` + scope resources

b-per:
Option 2 might actually help with allowing people to overwrite the env var on at the job level

Issue #99, opened 2022-11-29
Feature - Add ability to override env var values at the job level


b-per:
dbt Cloud allows overriding the environment variable values at the job level. Adding this capability would require modifying the existing `dbt_cloud_job` resource (or creating new resources?).

## Endpoints and payloads
I have some working code in Python but not in Go. Here are the specs for the CRUD operations.

### Get
- HTTP Method: `GET`
- url: `"{base_url}/api/v3/accounts/{account_id}/projects/{project_id}/environment-variables/job/?job_definition_id={job_id}"`

### Create
- HTTP Method: `POST`
- url: `"{base_url}/api/v3/accounts/{account_id}/projects/{project_id}/environment-variables/"`
- payload:
```
{
    "name": "<env_var_name>",
    "type": "job",
    "display_value": null,
    "job_definition_id": <job_id>,
    "id": null,
    "project_id":  <project_id>,
    "account_id":  <account_id>,
    "raw_value": "<value_for_the_override>"
}
```

### Update
- HTTP Method: `POST`
- url: `"{base_url}/api/v3/accounts/{account_id}/projects/{project_id}/environment-variables/{env_var_override_id}/"`
- payload:
```
{
    "name": "<env_var_name>",
    "type": "job",
    "display_value": null,
    "job_definition_id": <job_id>,
    "id": <env_var_override_id>,
    "project_id": <project_id>,
    "account_id": <account_id>,
    "raw_value": "<value_for_the_override>"
}
```

### Delete
- HTTP Method: `DELETE`
- url: `"{base_url}/api/v3/accounts/{account_id}/projects/{project_id}/environment-variables/{env_var_override_id}/"`
  -  where `env_var_override_id` is the `id` of the existing **override** (gotten from the `GET`)

b-per:
The endpoints being similar (but slightly different) to the `dbt_cloud_environment_variable` resource, and different than the `dbt_cloud_job` one, would it mean that we would create a new resource for those overrides?

Would calling it `dbt_cloud_environment_variable_job_override` make sense?

PR #225, opened 2023-11-22
Release 0.2.16

b-per:
## Changes

- #99 - add the resource `environment_variable_job_override` to allow environment variable override in jobs
- Update the go version and packages versions

## Fixes

- #221 - removing the value for an env var scope was not removing it in dbt Cloud
- add better messages and error handling for jobs


b-per:
Passing CI 🚀

PR #237, opened 2024-02-26
Prepare release 0.2.21

b-per:
- Add info about `versionless` dbt environment (Private Beta)
- Flag `fetch_deploy_key` as deprecated for `dbtcloud_repository`. The key is always fetched for the genetic git clone approach


b-per:
CI passed!

Issue #241, opened 2024-03-19
`dbtcloud_environment_variable` doesn't work when updating values without destroying/recreating them


b-per:
## Describe the bug

Some recent changes in the API changed the behaviour of `PUT` on the `bulk` endpoint.
Checking with the team if we want this to be updated on the Terraform side or on the API side.

Issue #240, opened 2024-03-18
Remove `custom_branch_only` from the triggers config for jobs


b-per:
## Describe the bug
`custom_branch_only` has been deprecated for jobs but some people still have it in their configs

Do the following to update the behaviour:

- [ ] make `custom_branch_only` not mandatory
- [ ] modify the logic so that people with `custom_branch_only=true` pass the value `false` anyway (as `true` is not a valid value anymore). A value of `true` should not trigger a new Terraform change every time
- [ ] update docs to mention that `custom_branch_only` has been deprecated

PR #242, opened 2024-03-20
Preparing new release with bug fixes and docs updates

b-per:
## Changes

- [#240](https://github.com/dbt-labs/terraform-provider-dbtcloud/issues/240) Add notice of deprecation for `triggers.custom_branch_only` for jobs and update logic to make it work even though people have it to true or false in their config. We might raise an error if the field is still there in the future.
- Update diff calculation for Extended Attributes, allowing strings which are not set with `jsonencode()`
- [#241](https://github.com/dbt-labs/terraform-provider-dbtcloud/issues/241) Force recreation of env vars when values change to work with the recent changes in the dbt Cloud API

## Documentation

- Add list of permission names and permission codes in the docs of the `service_token` and `group`
- Add info in `dbtcloud_repository` about the need to also create a `dbtcloud_project_repository`


b-per:
Passing CI 🚀 
(CI was failing before for env vars)

Issue #192, opened 2023-09-18
Issue when linking GitLab using Service Tokens


b-per:
When using service tokens to call the API to link a GitLab repo natively with dbt Cloud the connection is not set up correctly and require manually reconnecting the repo.

The same config works when using a user token.

This has been raised to the engineering team and should be automatically fixed without any action on this repo once the changes are done in dbt Cloud.

kcd83:
Thanks for reporting. This is still an issue. 

When diagnosing it I found the API for projects and/or repositories seems to show a little evidence of the problem

```
        "remote_backend": "gitlab",
        "git_clone_strategy": "deploy_token",
        "deploy_key_id": null,
        "repository_credentials_id": null,
```
Will not work, however setting this up as user in the UI I do see `repository_credentials_id` is set.

```
        "remote_backend": "gitlab",
        "git_clone_strategy": "deploy_token",
        "deploy_key_id": null,
        "repository_credentials_id": 6760,
```
The latter works fine

b-per:
Hi @kcd83 

Yes, this is still a problem, and the dbt Cloud team is aware of it, but there is nothing we can do just on the Terraform side for now. So, I keep this one as "blocked".

victorosetrov:
Hello @b-per and @adrianburusdbt 
following up on terraform-provider-dbtcloud#192
I can see It was closed 2026-06-03 as "completed," citing an expected dbt Cloud platform-side fix, but I have an enterprise customer reproducing the exact same error live as of last week, and the guard is still present in the current provider code (resource.go).

Could you help confirm: did the platform fix for GitLab actually ship, or was the issue closed before it landed? 
Trying to give the customer an accurate answer rather than relying on the closed status. 
Happy to share the full repro if useful.
Thank you

b-per:
I didn't close it so I am not sure, but to my knowledge this wasn't fixed on the dbt Cloud/platform side.

PR #244, opened 2024-03-25
docs: example - how to fail during apply when a service token is used…

matthiasng:
While migrating our dbt resoruces to terraform i encounter the gitlab integration issue #192 where you cant create a working repository in dbt with a service token. A big problem for me is that we will use the module I'm writing to create new dbt environments/jobs/projects in the future, but the apply stage doesn't throw an error when a service token is used.

So here is a small example of how to throw an error during the apply phase when a service token is used to create a Gitlab repository.

### DIFF
--- a/docs/resources/repository.md
+++ b/docs/resources/repository.md
@@ -64,6 +64,7 @@ resource "dbtcloud_repository" "github_repo_other" {
 
 ### repo cloned via the GitLab integration
 # as of 15 Sept 2023 this resource requires using a user token and can't be set with a service token - CC-791
+# for better clarity, you can add a null_resource with precondition to throw an error during apply if the repository is created with a service token
 resource "dbtcloud_repository" "gitlab_repo" {
   project_id         = dbtcloud_project.dbt_project.id
   remote_url         = "<gitlab-group>/<gitlab-project>"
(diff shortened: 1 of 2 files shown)


b-per:
Thanks for that! 
I will get the docs updated in the next release with those examples.

Not being ultra familiar with `null_resource`, is it fair to say that for Terraform 1.4+ we could also use the new [`terraform_data` resource](https://developer.hashicorp.com/terraform/language/resources/terraform-data)?

```hcl
resource "terraform_data" "check_repository" {
  lifecycle {
    precondition {
      condition     = dbtcloud_repository.gitlab_repo.repository_credentials_id != 0
      error_message = "Repositories created with a service token are currently not working properly (https://github.com/dbt-labs/terraform-provider-dbtcloud/issues/192)"
    }
  }
}
```

matthiasng:
I just tested it and `terraform_data` seems to work similar to `null_resource` in this case.

b-per:
I am thinking a bit more about it and I actually could/should add this check in the provider code itself. If `repository_credentials_id = 0` then raise an error message that it might be due to using a service token.

matthiasng:
That would be even better. If you run the check right after creating the repository, you could also try to remove it so that no broken repository is left behind (which is the case with my workaround)

b-per:
Handled in the logic in #246

Issue #247, opened 2024-04-09
segmentation violation code=0x1 addr=0x0 pc=0xaab1f8


kpucynski:
## Describe the bug

If token is set thru env variable and variable is empty plugin is crashing with SIGSEGV.

## Error message
```
Planning failed. Terraform encountered an error while generating this plan.
|
│ Error: Plugin did not respond
│ 
│   with provider["registry.terraform.io/dbt-labs/dbtcloud"],
│   on providers.tf line 16, in provider "dbtcloud":
│   16: provider "dbtcloud" {
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin.(*GRPCProvider).ConfigureProvider call. The plugin logs may contain
│ more details.
|

Stack trace from the terraform-provider-dbtcloud_v0.2.23 plugin:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xaab1f8]
goroutine 25 [running]:
github.com/dbt-labs/terraform-provider-dbtcloud/pkg/dbt_cloud.NewClient(0x0, 0x0, 0x0)
	github.com/dbt-labs/terraform-provider-dbtcloud/pkg/dbt_cloud/client.go:96 +0x98
github.com/dbt-labs/terraform-provider-dbtcloud/pkg/provider.providerConfigure({0xc8ad14?, 0x6?}, 0xc000310600?)
	github.com/dbt-labs/terraform-provider-dbtcloud/pkg/provider/provider.go:154 +0x21c
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Provider).Configure(0xc0000950e0, {0xdb5ec8, 0xc000310600}, 0xc000455090)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.30.0/helper/schema/provider.go:306 +0x222
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ConfigureProvider(0xc000307b00, {0xdb5ec8?, 0xc000310180?}, 0xc000446c30)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.30.0/helper/schema/grpc_provider.go:605 +0x3a9
[...]

PR #248, opened 2024-04-15
Minor fixes

b-per:
## Fixes

- [#247](https://github.com/dbt-labs/terraform-provider-dbtcloud/issues/247) Segfault when the env var for the token is empty
- [Internal] Issue with `job_ids` required to be set going forward, even if it is empty


b-per:
CI passing now

PR #251, opened 2024-04-21
Implement muxing to start using the new Plugin Framework

b-per:
## Changes

- Implements muxing to allow both SDKv2 and Plugin Framework resources to work at the same time. This change a bit the internals but shouldn't have any regression.
- Move some resources / datasources to the plugin Framework
- Remove legacy `dbt_cloud_xxx` resources


b-per:
CI passed 🚀

PR #254, opened 2024-04-24
Release 0.3.3 - Add `dbtcloud_group_partial_permissions`

### LINKED ISSUE #250: More flexible project permission set assignments for groups
I find the current approach for assigning project permission sets to user groups very limiting. In my organization, we maintain several dbt Cloud project infrastructure for various teams and will continue to create new projects as business needs evolve.

We would love to be able to assign a new project to a pre-existing group without having to worry about all the other pre-existing permission sets configured for that group. This makes evolution easier, as we don't need to be aware of pre-existing permission sets when we create more and more projects.

## Describe alternatives you've considered
Maintain some complexity in our Terraform code where we keep track of teams at the outset and their various projects.

## Additional context
N/A

b-per:
## Changes
Fixes #250

[Experimental] Create a new resource called `dbtcloud_group_partial_permissions` to manage permissions of a single group from different resources which can be set across different Terraform projects/workspaces. 

The dbt Cloud API doesn't provide endpoints for adding/removing single permissions, so the logic in the provider is more complex than other resources. If the resource works as expected for the provider users we could create similar ones for "partial" notifications and "partial" license mappings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,15 @@
 
 All notable changes to this project will be documented in this file.
 
-## [Unreleased](https://github.com/dbt-labs/terraform-provider-dbtcloud/compare/v0.3.1...HEAD)
+## [Unreleased](https://github.com/dbt-labs/terraform-provider-dbtcloud/compare/v0.3.3...HEAD)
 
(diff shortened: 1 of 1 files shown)


b-per:
CI passed 🚀

PR #273, opened 2024-07-09
Add Beta `writable_environment_categories` to `dbtcloud_service_token`

chasewalden:


chasewalden:
CC: @b-per . Should be ready to go

This was a long one, so lmk if I missed anything before merge

pkg/framework/objects/service_token/model.go
```diff
@@ -0,0 +1,107 @@
...
+// 	diffEnv1, diffEnv2 := lo.Difference(
+// 		listGroup1Envs,
+// 		listGroup2Envs,
+// 	)
+// 	return group1.PermissionSet == group2.PermissionSet &&
+// 		group1.ProjectID == group2.ProjectID &&
+// 		group1.AllProjects == group2.AllProjects &&
+// 		len(diffEnv1) == 0 &&
+// 		len(diffEnv2) == 0
+// }
```
b-per, line 107: Do we want to keep this or can we delete it?

pkg/framework/objects/service_token/resource.go
```diff
@@ -0,0 +1,325 @@
...
+			},
+			"state": schema.Int64Attribute{
+				Description: "Service token state (1 is active, 2 is inactive)",
+				Optional:    true,
+				Computed:    true,
+				Default:     int64default.StaticInt64(1),
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"service_token_permissions": schema.SetNestedBlock{
```
b-per, line 90: That's great that you kept a `SetNestedBlock` for backward compatibility. 👍 

For new resources or new nested attributes for existing resources Hashicorp recommends to move from [Nested Blocks](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/blocks/set-nested) to [Nested Attributes ](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#nested-attribute-types) but I don't think that we should move the SDK resources to the new type as the configuration format changes and we would break existing config/modules.

pkg/framework/objects/service_token/resource.go
```diff
@@ -0,0 +1,325 @@
...
+				Description: "Service token secret value (only accessible on creation))",
+				Computed:    true,
+				Sensitive:   true,
+			},
+			"state": schema.Int64Attribute{
+				Description: "Service token state (1 is active, 2 is inactive)",
+				Optional:    true,
+				Computed:    true,
+				Default:     int64default.StaticInt64(1),
+			},
```
b-per, line 87: Not related to this change but it reminds me that I think we should maybe at some point get rid of all the `state` fields. We can't create "deleted" resources or modify them to be "deleted" so the parameter doesn't make sense anywhere.

But, yes, some people might have added it to their configs.

pkg/framework/objects/service_token/data_source.go
```diff
@@ -0,0 +1,138 @@
...
+							Description: "Whether or not to apply this permission to all projects for this service token",
+							Computed:    true,
+						},
+						// TODO(cwalden): Would this be better as a Set of Int64?
+						// TODO(cwalden): Add a validator to ensure that the project ID is set if `all_projects` is false
+						"project_id": schema.Int64Attribute{
+							Description: "Project ID to apply this permission to for this service token",
+							Computed:    true,
+						},
+						// TODO(cwalden): Add validator to ensure that this is configurable for the given `permission_set`
```
b-per, line 86: Same comment as for the `resource` side.

pkg/framework/objects/service_token/data_source.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Schema implements datasource.DataSource.
+func (st *serviceTokenDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"service_token_id": schema.Int64Attribute{
+				Required:    true,
+				Description: "The ID of the service token",
+				// TODO(cwalden): Deprecate this in favor of `id`
+				// DeprecationMessage: "Use `id` instead",
```
b-per, line 51: Yes, in the SDKv2, `id` was a "special" field, a string that was supposed to hold the ID of the resource. With the Framework we don't really need it and might think of whether we want to deprecate the field for all resources/datasources once everything has moved to the Framework.

I also don't think that right now we are consistent with what the resource ID is (whether `id` or `<resource_type>_id`. Not a huge deal, but not ideal.

pkg/framework/objects/service_token/resource_acceptance_test.go
```diff
@@ -1,4 +1,4 @@
-package resources_test
+package service_token_test
```
b-per, line 1: Can we add a test or update the existing one where we add some environment level restrictions?

pkg/framework/objects/service_token/data_source_acceptance_test.go
```diff
@@ -1,4 +1,4 @@
-package data_sources_test
+package service_token_test
```
b-per, line 1: Like for the resource, can we add a test or update the existing one where we add some environment level restrictions?

docs/data-sources/service_token.md
```diff
@@ -17,20 +17,25 @@ description: |-
...
 
 Read-Only:
 
-- `all_projects` (Boolean)
-- `permission_set` (String)
-- `project_id` (Number)
+- `all_projects` (Boolean) Whether or not to apply this permission to all projects for this service token
+- `permission_set` (String) Set of permissions to apply
+- `project_id` (Number) Project ID to apply this permission to for this service token
+- `writable_environment_categories` (Set of String) What types of environments to apply Write permissions to.
```
b-per, line 37: Can we update [the example file](https://github.com/dbt-labs/terraform-provider-dbtcloud/blob/main/examples/resources/dbtcloud_service_token/resource.tf) + regenerate docs so that we show an example of how people can set restrictions at the env level?

pkg/dbt_cloud/environment_category.go
```diff
@@ -0,0 +1,26 @@
...
+	// Staging is the category for staging environments
+	Staging EnvironmentCategory = "staging"
+	// Production is the category for production environments
+	Production EnvironmentCategory = "production"
+	// Other is the category for other environments
+	Other EnvironmentCategory = "other"
+)
+
+// EnvironmentCategories is a list of all possible environment categories
+var EnvironmentCategories = []EnvironmentCategory{
```
b-per, line 20: Thanks for doing this. We should do it more in other parts of the code as well. So many places are not DRY today.

b-per (changes requested):
Thanks @chasewalden for doing all those changes 🙌 

I added a few comments, some of them are just informative and do not require a code change.

In a nutshell, before merging it we could:
- update the tests so that we test the new feature
- update the docs with an example of how we set env level restrictions
- check whether we want to keep the `TODO(cwalden)` in the code or remove those and maybe track future changes somewhere else

pkg/framework/objects/service_token/resource.go
```diff
@@ -0,0 +1,325 @@
...
+
+		_, err = st.client.UpdateServiceToken(svcTokID, *svcTok)
+		if err != nil {
+			resp.Diagnostics.AddError("Unable to update the service token", err.Error())
+			return
+		}
+	}
+
+	// Because we can't compare the service token permissions directly, we need to update them every time
+	// TODO(cwalden): Add a way to compare the service token permissions?
```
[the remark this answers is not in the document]
chasewalden: Reason I mentioned this is due to it being the behavior in the SDKv2 version. The equivalent code was wrapped in a `if d.HasChanged("...") { }` block. I was aiming for as close to 1:1 parity

pkg/framework/objects/service_token/resource.go
```diff
@@ -0,0 +1,325 @@
...
+							Required:    true,
+							Validators: []validator.String{
+								stringvalidator.OneOf(dbt_cloud.PermissionSets...),
+							},
+						},
+						"all_projects": schema.BoolAttribute{
+							Description: "Whether or not to apply this permission to all projects for this service token",
+							Required:    true,
+						},
+						// TODO(cwalden): Would this be better as a Set of Int64?
```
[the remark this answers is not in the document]
chasewalden: I thought a Set could make more sense: `all_projects` could be set in one block, but unset in another with a specific `project_id`. What would the expected behavior be in this scenario? 

I agree with keeping API parity though. Could be worth passing feedback up to the API team

pkg/framework/objects/service_token/model.go
chasewalden: More-or-less duplicated the code from `group/model.go`. I was unsure if I was going to need this during the refactor, but it appears to be a no. I will delete.

pkg/provider/sdk_provider.go
```diff
@@ -36,17 +36,17 @@ func SDKProvider(version string) func() *schema.Provider {
...
+				"dbtcloud_project":               data_sources.DatasourceProject(),
+				"dbtcloud_environment_variable":  data_sources.DatasourceEnvironmentVariable(),
+				"dbtcloud_snowflake_credential":  data_sources.DatasourceSnowflakeCredential(),
+				"dbtcloud_bigquery_credential":   data_sources.DatasourceBigQueryCredential(),
+				"dbtcloud_postgres_credential":   data_sources.DatasourcePostgresCredential(),
+				"dbtcloud_databricks_credential": data_sources.DatasourceDatabricksCredential(),
+				"dbtcloud_connection":            data_sources.DatasourceConnection(),
+				"dbtcloud_bigquery_connection":   data_sources.DatasourceBigQueryConnection(),
+				"dbtcloud_repository":            data_sources.DatasourceRepository(),
+				// "dbtcloud_service_token":            data_sources.DatasourceServiceToken(),
```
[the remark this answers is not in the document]
chasewalden: Oh. Good catch, missed these!

pkg/framework/objects/service_token/resource.go
chasewalden: I'll create a Task Ticket to make all `state` fields Optional + Deprecated. We can push a release without breaking anyone and give a timeline for when it will be removed.

pkg/framework/objects/service_token/data_source.go
chasewalden: Same response as for the `resource` side 😉

pkg/framework/objects/service_token/data_source.go
chasewalden: IMO, I'd prefer `id` over `<resource_type>_id`. It better aligns with the APIs. Based on other usage patterns in the APIs and Provider, `<resource_type>_id` implies a "foreign key" style relationship, while `id` is "primary key"

pkg/framework/objects/service_token/resource_acceptance_test.go
chasewalden: I will add. Can we check to make sure that the feature is enabled in the instance used by CI ?

pkg/framework/objects/service_token/data_source_acceptance_test.go
chasewalden: I will add. Can we check to make sure that the feature is enabled in the instance used by CI ?

pkg/dbt_cloud/environment_category.go
chasewalden: Definitely. I try to do little refactors like this when I come across it

pkg/framework/objects/service_token/resource.go
```diff
@@ -0,0 +1,325 @@
...
+						// TODO(cwalden): Add a validator to ensure that the project ID is set if `all_projects` is false
+						"project_id": schema.Int64Attribute{
+							Description: "Project ID to apply this permission to for this service token",
+							Optional:    true,
+						},
+						// TODO(cwalden): Add validator to ensure that this is configurable for the given `permission_set`
```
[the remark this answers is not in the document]
chasewalden: Given your reasoning, I am inclined to agree with 2. I am worried about possible issues with an old provider using new features. In this case, I doubt there will be a compatibility issue, but realistically it shouldn't be unreasonable for a user to need to upgrade to access new features/constraints

pkg/framework/objects/service_token/resource.go
chasewalden: Yes, I would have used a NestedAttribute here, but the SDKv2 version used a block. This was 100% for backwards compatibility. I aimed for 1:1 parity with the old resource.

pkg/framework/objects/service_token/resource.go
chasewalden: See TFPCLD-13 & TFPCLD-14

pkg/framework/objects/service_token/data_source.go
chasewalden: Lets have a chat on this. There are a few tricky things to contend with here.
1. SDKv2's built-in `id` is always a String, but we typically use Int64. Do we stick with Strings and always convert to/from Int64? Changing it is a breaking change so any user that relies on it being a String could face some issues.
2. How do we handle the removal of `service_token_id`?
3. Will we need/want to implement a StateUpgrader?

pkg/framework/objects/service_token/resource_acceptance_test.go
b-per: It is, because I implemented the env level for groups already and added tests for it.

The same flag is used for groups and service tokens.

chasewalden:
Lets fix the typo mentioned in #275 here as well.

pkg/framework/objects/service_token/resource.go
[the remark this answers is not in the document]
chasewalden: To fix with TFPCLD-16

pkg/framework/objects/service_token/resource.go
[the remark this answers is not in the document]
chasewalden: To fix with TFPCLD-16

pkg/framework/objects/service_token/resource.go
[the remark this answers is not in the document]
chasewalden: Mitigated after TFPCLD-17

pkg/framework/objects/service_token/resource.go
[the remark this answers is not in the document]
chasewalden: Resource level validation emits a warning

docs/data-sources/service_token.md
chasewalden: Done!

chasewalden:
Should have everything we need to Merge & Close. 

We can also close #275 with this as the fix to the typo is bundled in with this.

pkg/sdkv2/resources/service_token.go
b-per, line 1: We should delete this whole file, no?

b-per (changes requested):
Thanks @chasewalden for all the new changes.

I just left a few minor comments before we can merge it.

pkg/sdkv2/resources/service_token_acceptance_test.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 func TestAccDbtCloudServiceTokenResource(t *testing.T) {
+	t.Skip("Test deprecated")
```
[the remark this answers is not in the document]
chasewalden: Yes, you are right. I skipped this so that I could keep the file around while I ported the tests. This can be removed as it has been moved wholesale into the framework.

b-per (approved):

PR #275, opened 2024-07-12
fix typo in docs/resources/service_token

tanukifk:
I found this typo when I tried to import my service token resource.


chasewalden:
This is fixed as a part of #273 

Thanks for contributing!

PR #276, opened 2024-07-16
Add a new datasource for `dbtcloud_jobs` to retrieve multiple jobs

b-per:
- add new datasource + docs + tests
- fix typo in diagram
- refactor helper a bit


chasewalden:
Looks good to me. 

As a side-note, this does make me think we should eventually go through an clean up uses of `int` in favor of a fixed-width type (either `int32` or `int64`). Not a big deal for right now, but will help with unnecessary conversions back-and-forth and would prevent hard-to-debug issues between 32 & 64 bit machines in the future

chasewalden (approved):

PR #277, opened 2024-07-17
Add new `dbtcloud_users` datasource to get all users

b-per:
Today, some people want to set internal notifications based on email addresses.

One of the ways to handle it is to get the id of a user with the `dbtcloud_user` datasource but it fails if the email doesn't exist.

This new data source returns all users so people can do further filtering in HCL if need be. It handles pagination for people with more than 100 users.


b-per:
CI passed 🚀

PR #278, opened 2024-07-18
Deprecating `state` field

chasewalden:


chasewalden:
CC: @b-per 

Small change for the deprecation of the `state` attribute

pkg/sdkv2/data_sources/privatelink_endpoint.go
```diff
@@ -20,26 +20,27 @@ var privatelinkEndpointSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeString,
 		Computed:    true,
 		Description: "The CIDR range of the PrivateLink Endpoint",
 	},
-	"state": &schema.Schema{
+	"state": {
 		Type:        schema.TypeInt,
 		Computed:    true,
 		Description: "PrivatelinkEndpoint state should be 1 = active, as 2 = deleted",
+		Deprecated:  "Remove this attribute's configuration as it's no longer in use and the attribute will be removed in the next major version of the provider.",
```
b-per, line 43: This is from a datasource and is computed, so people won't be able to remove its configuration.
I am not too sure about how TF handles deprecation for Computed fields actually.

Do we just update the description for those and mention that the field is going to be deprecated, without the comment that people need to remove its configuration?

PR #291, opened 2024-08-28
attempt to fix flakiness with notifications

Topherhindman:
We've been seeing alot of flakiness with notifications related tests.  

Here's some examples:
https://github.com/dbt-labs/dbt-cloud/actions/runs/10599739892/job/29377047288?pr=12505#step:12:218
https://github.com/dbt-labs/dbt-cloud/actions/runs/10599739892/job/29380913981?pr=12505#step:12:208
https://github.com/dbt-labs/dbt-cloud/actions/runs/10599739892/job/29381607840?pr=12505#step:12:208
https://github.com/dbt-labs/dbt-cloud/actions/runs/10600515659/job/29379537697?pr=12506#step:11:217

I believe it might all stem from this, potentially.  Somewhat blindly attempting this. 

Ran the tests 3 times to try and better ensure this works and doesn't cause any unforeseen problems, and all 3 succeeded.

### DIFF
--- a/pkg/framework/objects/partial_notification/resource_acceptance_test.go
+++ b/pkg/framework/objects/partial_notification/resource_acceptance_test.go
@@ -30,6 +30,7 @@ func TestAccDbtCloudPartialNotificationResource(t *testing.T) {
 	resource.Test(t, resource.TestCase{
 		PreCheck:                 func() { acctest_helper.TestAccPreCheck(t) },
 		ProtoV6ProviderFactories: acctest_helper.TestAccProtoV6ProviderFactories,
+		CheckDestroy:             testAccCheckDbtCloudPartialNotificationDestroy,
 		Steps: []resource.TestStep{
 			{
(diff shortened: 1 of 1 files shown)


b-per:
I don't think that the suggested change will fix it.

What it would do is check that at the end of the test the resource is really destroyed. 

But the issues are that when we create a notification and do a `GET` for it just after, it is not returned by the API. And then, because of earlier failures there were dangling notifications that created different issues.

Should I just deactivate the notification tests for dbt Cloud PRs?

Topherhindman:
> I don't think that the suggested change will fix it.
> 
> What it would do is check that at the end of the test the resource is really destroyed.
> 
> But the issues are that when we create a notification and do a `GET` for it just after, it is not returned by the API. And then, because of earlier failures there were dangling notifications that created different issues.
> 
> Should I just deactivate the notification tests for dbt Cloud PRs?

Sure thats probably the best thing for now until it can be properly fixed

PR #294, opened 2024-09-02
Add datasources for projects and global connections

### LINKED ISSUE #288: Allow to retrieve i.a. connection and environment ID by their name
For `dbtcloud_project` we can do:

```
// or they can use project names
// the provider will raise an error if more than one project is found with the same name
data "dbtcloud_project" "test_project" {
  name = "My project name"
}
```

https://registry.terraform.io/providers/dbt-labs/dbtcloud/latest/docs/data-sources/project

It would be great to achieve the same for other named resources like connection or environment.

b-per:
Closes #288

### Changes

- Add a `dbtcloud_projects` (with an "s") datasource to return all the projects along with some information about the warehouse connections and repositories connected to those projects. Loops through the API in case there are more than 100 projects
  - Along with the `check` block, it can be used to check that there are no duplicate project names for example.
- Add a datasource for `dbtcloud_global_connection` with the same information as the corresponding resource
- Add a datasource for `dbtcloud_global_connections` (with an "s"), returning all the connections of an account along with details like the number of environments they are used in. This could be used to check that connections  don't have the same names or that connections are all used by projects.


b-per:
CI passed 🚀

PR #296, opened 2024-09-12
[CC-1769] - Make dbname for postgres, redshift a required field due to api updates

sidpatel13:
`dbname` is a required field that we do validation on in response but not on the request. We should have the same validation rules both ways. Changing this broke the following terraform test:
```
"developer_message":"Invalid request data: 1 validation error for
        PostAccountConnection\n__root__ -> dbname\n  none is not an allowed value
```

So updating this to pass in a db name for the terraform test.


MitchellBarker (approved):

b-per:
Thanks. Getting it merged to unlock everyone.
I will add another PR to make the field mandatory in the TF config for Postgres and Redshift

Issue #299, opened 2024-09-20
Terragrunt/Terraform: Error: Plugin did not respond


davidfuturelabs:
## Describe the bug

Error when trying import configuration already existent in terragrunt.

## Error message

xxx@xxx environments % terragrunt import 'dbtcloud_environment.environments["Development"]' 

dbtcloud_environment.environments["Development"]: Importing from ID "myid"...
dbtcloud_environment.environments["Development"]: Import prepared!
  Prepared dbtcloud_environment for import
dbtcloud_environment.environments["Development"]: Refreshing state... [id=myid]
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details.
╵

Stack trace from the terraform-provider-dbtcloud_v0.3.16 plugin:

panic: runtime error: index out of range [1] with length 1

goroutine 12 [running]:
github.com/dbt-labs/terraform-provider-dbtcloud/pkg/sdkv2/resources.resourceEnvironmentRead({0x101758498?, 0x1400056e690?}, 0x1400063eb00, {0x101741420?, 0x1400043d2c0})
	github.com/dbt-labs/terraform-provider-dbtcloud/pkg/sdkv2/resources/environment.go:188 +0xad0
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0x140002fc380, {0x101758498, 0x1400056e690}, 0xd?, {0x101741420, 0x1400043d2c0})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema/resource.go:823 +0xe8
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0x140002fc380, {0x101758498, 0x1400056e690}, 0x140004b2750, {0x101741420, 0x1400043d2c0})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema/resource.go:1117 +0x430
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadResource(0x1400011b8a8, {0x101758498?, 0x1400056e540?}, 0x140000a8540)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema/grpc_provider.go:708 +0x5ac
[...]

PR #302, opened 2024-09-26
Add better error handling when splitting IDs

b-per:
Add better error handling instead of panicking when IDs can't be split properly.

Related to #299


b-per:
CI passed 🚀

PR #321, opened 2024-12-06
chore: Migrate Azure DevOps Project Datasource from SDKv2 to the plugin framework

csquire:
Migrates the `dbtcloud_azure_dev_ops_project` datasource from the legacy SDKv2 to the Terraform Plugin Framework.


b-per (commented):
It looks that there was no test for the SDKv2 datasource, but would it make sense to add a test of the new datasource that we would run only in Dev (I don't think that we would be able to test in CI as we might not have an ADO integration in place)

chasewalden:
> It looks that there was no test for the SDKv2 datasource, but would it make sense to add a test of the new datasource that we would run only in Dev (I don't think that we would be able to test in CI as we might not have an ADO integration in place)

I think that's the challenge. We're getting other team members ramped in the codebase, and since there were no tests, it might be tricky to retroactively add in tests, especially if there's not ADO integration available. We have a ticket to add further testing where there seem to be gaps

chasewalden:
@b-per , I'm going to get this merged to keep the rest of the team on track. I will add acc tests separately

chasewalden (approved):

PR #325, opened 2024-12-12
chore: Migrate license map resource from SDKv2 to plugin framework

csquire:
Migrates the `dbtcloud_license_map` resource from the legacy SDKv2 to the Terraform Plugin Framework.


docs/resources/license_map.md
```diff
@@ -45,7 +45,7 @@ resource "dbtcloud_license_map" "it_license_map" {
 
 ### Read-Only
 
-- `id` (String) The ID of this resource.
```
chasewalden, line 48: I don't know if there are potential breaking changes for changing the datatype here. I know that the ID *is* a number, but SDKv2 treats all IDs as strings. I also know there is some implicit type coercion that TF does. Although I'm tempted to change this (in other resources/data-sources too), I think we should stay conservative and minimize datatype changes for the time being.

chasewalden: @b-per If you have thoughts here, let me know

chasewalden (commented):
Overall looks good. Want to check in on the datatype change on `id`, but otherwise I am happy to approve once we can touch base on that.

docs/resources/license_map.md
csquire: I did wonder about this as well, since the id field wasn't defined explicitly under SDKv2. The switch to a Number is due to use the model that was initially created for `partial_license_map`. This note regarding type conversions gives me some level of confidence: https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion 

If I apply with the old provider where the id is stored as a string, I get the following in the state:
```
terraform state show dbtcloud_license_map.test_license_map                                      
# dbtcloud_license_map.test_license_map:
resource "dbtcloud_license_map" "test_license_map" {
    id                         = "35218731827208"
    license_type               = "read_only"
    sso_license_mapping_groups = [
        "group1",
        "group2",
    ]
}
```

Switching to this branch and running again, the plan/apply come back clean:
```
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

Destroying and reapplying with the new provider results in a Number in the state:
```
terraform state show dbtcloud_license_map.test_license_map                                 
# dbtcloud_license_map.test_license_map:
resource "dbtcloud_license_map" "test_license_map" {
    id                         = 35218731827208
    license_type               = "read_only"
    sso_license_mapping_groups = [
        "group1",
        "group2",
    ]
}
```

b-per: Agree, I am pretty sure that TF won't break if ID changes from an int stored as string to an int stored as int, and it will be transparent for folks using those IDs.

The fact that it was a string in SDKv2 was enforced by the framework, but now we can give it a proper type

chasewalden (approved):

PR #326, opened 2024-12-13
Pagination support for service_token_permissions

### LINKED ISSUE #280: Error when updating token resource in 0.3.10
## Describe the bug
In the new provider version 0.3.10 updating `dbtcloud_service_token` resource is failing.

## Error message
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ dbtcloud_service_token.tokens["token_name"],
│ provider "provider[\"registry.terraform.io/dbt-labs/dbtcloud\"]" produced
│ an unexpected new value: .service_token_permissions: planned set element
│ cty.ObjectVal(map[string]cty.Value{"all_projects":cty.False,
│ "permission_set":cty.StringVal("developer"),
│ "project_id":cty.NumberIntVal(3597),
│ "writable_environment_categories":cty.SetVal([]cty.Value{cty.StringVal("all")})})
│ does not correlate with any element in actual.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

## Resource configuration
Service tokens are configured with the loop.
Affected token have around 50 project IDs assigned. 
```
resource "dbtcloud_service_token" "tokens" {
  for_each = local.tokens

  name = each.key

  dynamic "service_token_permissions" {
    for_each = each.value
    content {
      permission_set = "developer"
      all_projects   = false
      project_id     = service_token_permissions.value
    }
  }

  dynamic "service_token_permissions" {
    for_each = each.value
    content {
      permission_set = "git_admin"
      all_projects   = false
      project_id     = service_token_permissions.value
    }
  }

  dynamic "service_token_permissions" {
    for_each = each.value
    content {
      permission_set = "job_admin"
      all_projects   = false
      project_id     = service_token_permissions.value
    }
  }
}
```

## Expected behavior
Token updated in place.

- Installed dbt-labs/dbtcloud v0.3.10 (self-signed, key ID 644DCEA333782266)

## Additional context
Downgrade to version 0.3.8 solves the problem and update is passing.

ashtonwalden:
Added pagination support for service_token_permissions.

Resolves #280

### DIFF
--- a/pkg/dbt_cloud/paginate.go
+++ b/pkg/dbt_cloud/paginate.go
@@ -15,6 +15,11 @@ type Response struct {
 	Extra Extra `json:"extra"`
 }
 
+type RawResponse struct {
+	Data  []json.RawMessage `json:"data"`
+	Extra Extra             `json:"extra"`
+}
+
 type Extra struct {
 	Pagination Pagination `json:"pagination"`
 }
@@ -40,19 +45,19 @@ func (c *Client) GetEndpoint(url string) ([]byte, error) {
 	return resp, err
 }
 
-func (c *Client) GetData(url string) []any {
+func (c *Client) GetRawData(url string) ([]json.RawMessage, error) {
 
 	// get the first page
 	jsonPayload, err := c.GetEndpoint(url)
 	if err != nil {
-		log.Fatal(err)
+		return nil, err
 	}
 
-	var response Response
+	var response RawResponse
 
 	err = json.Unmarshal(jsonPayload, &response)
 	if err != nil {
-		log.Fatal(err)
+		return nil, err
 	}
 
 	allResponses := response.Data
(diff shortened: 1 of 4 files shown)


pkg/dbt_cloud/paginate.go
```diff
@@ -90,7 +95,23 @@ func (c *Client) GetData(url string) []any {
...
+
+func (c *Client) GetData(url string) []any {
+	rawData, err := c.GetRawData(url)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	allData := make([]any, len(rawData))
+	for i, data := range rawData {
+		err := json.Unmarshal(data, &allData[i])
```
b-per, line 109: The acceptance test is now passing, which is great 👍 , but I am just wondering why we need to add a new unmarshalling step to the process

ashtonwalden: Small performance tweak. We added that to try and prevent the data from needing to be marshalled and unmarshaled again by the caller (ie: 161-163 or 180-182). Might make sense to try and swap some of the other client calls over to the GetRawResponse method.

csquire: @b-per The magic lies in Go's `json.RawMessage` type, so the actual data doesn't get unmarshalled just to read the pagination info.

csquire (approved):
LGTM 🚢 
Maybe we want to rename the PR title?

PR #327, opened 2024-12-16
WIP: Migrate Job DataSource & Resource

chasewalden:


chasewalden:
@b-per Agreed. This is going to be given extra attention. Trying to do this incrementally. First was migrating the schemas and ensuring the docs aren't (too) different. There are some plan modifiers that need to be written for some of the custom logic (i.e. some mutually exclusive attributes, recreate on certain trigger behavior, etc). Datasource will be next followed by the Resource CRUD methods. I'll drop in extra unit/acc tests where I can.

PR #338, opened 2025-01-27
Add ability to configure linting on CI jobs

csquire:
Resolves https://github.com/dbt-labs/terraform-provider-dbtcloud/issues/310 

Adds fields to the `dbtcloud_job` resource in support of linting. Also just a bit of code cleanup by switching some internal job-related struct fields to camel-case instead of snake-case.

- `run_lint`: Whether the CI job should lint SQL changes. Defaults to `false`.
- `errors_on_lint_failure`: Whether the CI job should fail when a lint error is found. Only used when `run_lint` is set to `true`. Defaults to `true`.

Note that the defaults match dbt Cloud's default behavior.

Example:
```
resource "dbtcloud_job" "ci_job" {
  environment_id = dbtcloud_environment.ci_environment.environment_id
  execute_steps = [
    "dbt build -s state:modified+ --fail-fast"
  ]
  generate_docs            = false
  deferring_environment_id = dbtcloud_environment.prod_environment.environment_id
  name                     = "CI Job"
  num_threads              = 32
  project_id               = dbtcloud_project.dbt_project.id
  run_generate_sources     = false
  run_lint                 = true
  errors_on_lint_failure   = true
  triggers = {
    "github_webhook" : true
    "git_provider_webhook" : true
    "schedule" : false
    "on_merge" : false
  }
  # this is the default that gets set up when modifying jobs in the UI
  # this is not going to be used when schedule is set to false
  schedule_days = [0, 1, 2, 3, 4, 5, 6]
  schedule_type = "days_of_week"
}
```

### DIFF
--- a/pkg/dbt_cloud/job.go
+++ b/pkg/dbt_cloud/job.go
@@ -9,15 +9,15 @@ import (
 )
 
 type JobTrigger struct {
-	Github_Webhook     bool `json:"github_webhook"`
+	GithubWebhook      bool `json:"github_webhook"`
 	Schedule           bool `json:"schedule"`
 	GitProviderWebhook bool `json:"git_provider_webhook"`
 	OnMerge            bool `json:"on_merge"`
 }
 
 type JobSettings struct {
-	Threads     int    `json:"threads"`
-	Target_Name string `json:"target_name"`
+	Threads    int    `json:"threads"`
+	TargetName string `json:"target_name"`
 }
 
 type scheduleDate struct {
@@ -44,7 +44,7 @@ type JobResponse struct {
 }
 
 type JobExecution struct {
-	Timeout_Seconds int `json:"timeout_seconds"`
+	TimeoutSeconds int `json:"timeout_seconds"`
 }
 
 type JobCompletionTrigger struct {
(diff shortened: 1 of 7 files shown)


pkg/dbt_cloud/job.go
```diff
@@ -9,15 +9,15 @@ import (
 )
 
 type JobTrigger struct {
-	Github_Webhook     bool `json:"github_webhook"`
```
b-per, line 12: Did you update all of those by hand or used some tool for it? I have been writing new code with CamlCase but didn't touch most of the code that was written before I got involved.

pkg/dbt_cloud/job.go
```diff
@@ -59,26 +59,28 @@ type JobCompletionTriggerCondition struct {
...
 type Job struct {
 	ID                     *int                  `json:"id"`
-	Account_Id             int                   `json:"account_id"`
-	Project_Id             int                   `json:"project_id"`
-	Environment_Id         int                   `json:"environment_id"`
+	AccountId              int                   `json:"account_id"`
+	ProjectId              int                   `json:"project_id"`
+	EnvironmentId          int                   `json:"environment_id"`
 	Name                   string                `json:"name"`
+	DbtVersion             *string               `json:"dbt_version"`
```
b-per, line 66: Same here. Did you sort them by alphabetical order or are you using a tool for it? Historically, I have been trying to group new fields together but it also makes sense to have those in alphabetical order.

pkg/sdkv2/resources/job_acceptance_test.go
```diff
@@ -236,6 +236,98 @@ func TestAccDbtCloudJobResourceTriggers(t *testing.T) {
...
+					projectName,
+					environmentName,
+					false,
+					true,
+				),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDbtCloudJobExists("dbtcloud_job.ci_job"),
+					resource.TestCheckResourceAttr("dbtcloud_job.ci_job", "name", jobName),
+					resource.TestCheckResourceAttr("dbtcloud_job.ci_job", "run_lint", "false"),
+					resource.TestCheckResourceAttr("dbtcloud_job.ci_job", "errors_on_lint_failure", "true"),
```
b-per, line 262: Maybe a bigger discussion for the future, but I read somewhere that we actually might not need to check attributes that are not `Computed`. The testing framework actually checks that all the non Computed values are OK based on the config (e.g. it makes sure that a `plan` after an `apply` would not raise any diff). I know that this is not the way that the provider has been written so far though.

b-per (approved):
Thanks @csquire 
I left some comments but more for my own understanding.

LGTM
The failing test is with another resource and might be related  to a transient issue with Tableau at that time, it is not related to your change.

Let's make sure that we mention the bugfix in the changelog when we do the new release.

pkg/dbt_cloud/job.go
csquire: I used IntelliJ (GoLand) to do the rename, so it updates all references as well.

pkg/dbt_cloud/job.go
csquire: The ordering I did by hand using IntelliJ to move the lines up/down (cmd-shift up/down).

pkg/sdkv2/resources/job_acceptance_test.go
csquire: That makes sense! Let me read up on it and maybe I can remove those!

PR #340, opened 2025-01-28
(CI) Add concurrency limit on acceptance tests

ashtonwalden:
- Added concurrency limit for running acceptance tests
- Added test run on merge to main


csquire (approved):
I like this. Would be nice to have the acceptance tests run concurrently, but I believe there are some of them that would stomp on each other.

PR #355, opened 2025-03-19
Migrate bigquery_credential source and resource to framework

adrianburusdbt:
Migrate bigquery_credential source and resource to framework


pkg/framework/objects/bigquery_credential/resource.go
```diff
@@ -0,0 +1,323 @@
...
+
+		resp.Diagnostics.AddError(
+			"Error reading Bigquery credential",
+			"Could not read Bigquery credential ID "+state.ID.ValueString()+": "+err.Error(),
+		)
+		return
+	}
+
+	// Refresh state values
+	state.CredentialID = types.Int64Value(int64(*credential.ID))
```
vladmdbt, line 155: Don't we need to set state.ID here? I saw on the other resources that the state.ID is set in the Read()

vladmdbt: In the starburst_credential resource the state.ID seems not to be set.

pkg/framework/objects/bigquery_credential/resource.go
```diff
@@ -0,0 +1,323 @@
...
+	// Get current state
+	var state BigqueryCredentialResourceModel
+	diags = req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectID := int(plan.ProjectID.ValueInt64())
+	credentialID := int(state.CredentialID.ValueInt64())
```
vladmdbt, line 192: plan instead of state here?
int(**_plan_**.CredentialID.ValueInt64())

pkg/framework/objects/bigquery_credential/resource.go
```diff
@@ -0,0 +1,323 @@
...
+			"Unexpected Import Identifier",
+			fmt.Sprintf(
+				"Could not convert credential_id to integer. Got: %q",
+				idParts[1],
+			),
+		)
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(
```
vladmdbt, line 308: Shouldn't we fetch here the other fields related to this resource and append them to the state?
	IsActive     types.Bool   `tfsdk:"is_active"`
	Dataset      types.String `tfsdk:"dataset"`
	NumThreads   types.Int64  `tfsdk:"num_threads"`
	
Like in this case: https://github.com/dbt-labs/terraform-provider-dbtcloud/blob/55e11fcc477f23b310e2f08234db1b0f1fa05823/pkg/framework/objects/starburst_credential/resource.go#L319

PR #375, opened 2025-03-31
SIGN-37 Isolate conformance tests

victorasu:


vladmdbt (approved):

Makefile
```diff
@@ -18,9 +21,12 @@ doc:
...
 
 test: deps
-	go test -mod=readonly -count=1 ./...
+	go test -run $(EXCLUDE_CONF) -mod=readonly -count=1 ./...
 
 test-acceptance: deps
+	TF_ACC=1 go test -v -run $(EXCLUDE_CONF) -mod=readonly -count=1 -p 1 -parallel 10 ./...
+
+test-conformance: deps
 	TF_ACC=1 go test -v -mod=readonly -count=1 -p 1 -parallel 10 ./...
```
VanTudor, line 30: IIRC, we decided offline that `test-conformance` will only test conformance tests, and not the others. In other words, this should use the inverse of the `EXCLUDE_CONF` regexp

VanTudor: Let's also call this in our CI/CD pipeline, otherwise, this PR would effectively skip conformance tests. 

Question: do you know if Cloud's workflow calls make `test-acceptance` or actually invokes our workflow somehow? I'm thinking that we don't want to enable conformance tests on their side.

VanTudor (changes requested):

Makefile
b-per: The dbt Cloud PR is calling `make test-acceptance` on their side: https://github.com/dbt-labs/dbt-cloud/blob/a11f22001020fc8b6dfb6f5768b0aba5d6218d61/.github/workflows/main.yml#L1715

victorasu: Benoit is right, they’re specifically calling `make test-acceptance` after checking out the repo. We should be safe on that dependency.

I must have forgotten about the test-conformance attribute, I’ll be adding that as well. Good catch!!

VanTudor (approved):
Nice! Lgtm

Makefile
VanTudor: Thanks @b-per! I think we're safe with merging this, then

PR #379, opened 2025-04-02
change test to use same type cred/conn.

fernando-toledo:
Context:
- We want to add a constraint to prevent connections from having credentials of a different type associated with them in dbt-cloud repo.
- For example, a Snowflake connection type should not have a BigQuery credential type associated.
- After adding this constraint to dbt-cloud and running Terraform tests, some tests failed due to this violation, specifically: TestAccDbtCloudEnvironmentResourceConnection.


adrianburusdbt:
Hey @fernando-toledo, do you know why it fails currently, considering that the resource dbtcloud_environment is not specific to snowflake? e.g. it should work with bigquery too, right?
I updated the branch and moved your changes to the new location of the resource, so it should be fine, just curious of the above

adrianburusdbt (approved):

PR #403, opened 2025-05-02
Support Type setting of dbtcloud_project resource

trouze:
Resolves issue https://github.com/dbt-labs/terraform-provider-dbtcloud/issues/398 by adding support for type in the dbtcloud_project resource.

### DIFF
--- a/pkg/framework/objects/project/schema.go
+++ b/pkg/framework/objects/project/schema.go
@@ -41,6 +41,10 @@ var datasourceSchema = datasource_schema.Schema{
 						Computed:    true,
 						Description: "Subdirectory for the dbt project inside the git repo",
 					},
+					"type": datasource_schema.Int64Attribute{
+						Computed:    true,
+						Description: "The type of dbt project (default or hybrid)",
+					},
 					"created_at": datasource_schema.StringAttribute{
 						Computed:    true,
 						Description: "When the project was created",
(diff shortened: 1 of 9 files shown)


vladmdbt:
Awesome contribution! I was about to test the project data source/ resource more thoroughly after we moved it from sdkv2 so the timing is perfect. Feel free to merge whenever you want. :)

pkg/framework/objects/project/schema.go
```diff
@@ -204,5 +212,13 @@ var resourceSchema = resource_schema.Schema{
 				stringplanmodifier.UseStateForUnknown(),
 			},
 		},
+		"type": resource_schema.Int64Attribute{
+			Description: "The type of dbt project (0=default or 1=hybrid)",
+			Optional:    true,
+			Computed:    true,
```
b-per, line 218: Do we want `Computed: true` or to set the default value to 0?

I think that setting the default value to 0 makes more sense to me. 

This is not a huge difference but the main one is with `Computed`, we wouldn't trigger a change if someone change the project from a "default" one to an "hybrid" one in the UI. I don't even know if this is technically possible for this particular use case, but if it is, I think that the expected behaviour from a `plan`/`apply` after someone changed the project would be to make it again a "default" one.

vladmld: I noted this down and we'll see if we add this change to the 1.0.0 release.

PR #420, opened 2025-05-13
Sign 111 semantic layer bigquery

teodoravochin-dbtlabs:


pkg/framework/objects/semantic_layer_credential/resource.go
```diff
@@ -19,6 +19,7 @@ func SnowflakeSemanticLayerCredentialResource() resource.Resource {
 	return &snowflakeSemanticLayerCredentialResource{}
```
vladmld, line 19: I would rename this file from `resource.go` to `snowflake_sl_credential_resource.go` to be consistent with `bigquery_sl_credential_resource.go`. And as we add others we keep the same naming convetion. What do you think?

Also the acceptance files to bigquery_sl_credential_resource_acceptance

vladmld:
If you merge main into the branch you can nou use `changie` to update the changelog. 

in the root folder of the project run `changie new`, select a type (I think `Changes` for this case), add what the changes are it will generate `*.yml` file with the changes, commit that and you're good to go. 🚀

vladmld (approved):

PR #425, opened 2025-05-16
Roll back to v0.3.26

victorasu:
This PR is a failsafe in case the release process generates critical failure. This would mean that we need to do a roll forward, due to the problematic version having to be superseded by a working one.


victorasu:
**DO NOT MERGE, ONLY USE IN CASE OF DISASTER**

vladmld:
Great idea!

PR #434, opened 2025-05-23
Fixed optional params for environment

adrianburusdbt:
An issue was reported that an error `Error: Provider produced inconsistent result after apply` was triggered on `dbtcloud_environment` resource in a scenario where optional parameters were not provided.
This fixes the way optional (non-computed) params are overwritten in case they are unknown.


pkg/framework/objects/environment/SIGN_136_test.go
victorasu, line 1: I do think we should rename this to something more generic.

victorasu (approved):

pkg/framework/objects/environment/SIGN_136_test.go
adrianburusdbt: I kind of did this on purpose to track it easier, but I will rename it and add a comment instead

PR #435, opened 2025-05-23
Sign 128 sematic layer creds redshift

teodoravochin-dbtlabs:


victorasu:
Please make sure to run `make doc` after creating new resources/data sources.

PR #456, opened 2025-06-04
Fix orphaned semantic layer configuration

phildier:
When creating a semantic layer configuration on a project via terraform, the project isn't being update with the configuration ID, causing it to not appear in the UI and presumably being orphaned.  This adds code to fetch the project and update it with the configuration ID so that it appears in the UI.

I determined this was necessary by following the network request that are made in the UI when creating a semantic config.  I noticed that a POST was made to 2 endpoints when creating a semantic layer configuration.  One to create the configuration, and one to update the project with the configuration ID.

### DIFF
--- a/.changes/unreleased/Fixes-20250604-114343.yaml
+++ b/.changes/unreleased/Fixes-20250604-114343.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: Fixes orphaned semantic layer configuration not showing up in project
+time: 2025-06-04T11:43:43.043953356-05:00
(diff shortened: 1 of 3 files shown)


teodoravochin-dbtlabs:
Thanks for identifying this and for opening the PR. The changes look good. I checked the APIs called in the cloud UI after adding a semantic layer configuration, and it is indeed making an update to the project so that a configuration ID is added to the project.

Regarding the failing build that you have, you should run the command 'changie new' in the terminal, complete the generated file with a short description about what this PR is about, commit the file and push it. The builds should be successful after this.

PR #464, opened 2025-06-16
SIGN-177 Fix for Branch link in IDE redirects to https://github.com/None/tree/main on projects that were created with terraform

vladmld:
- Added `RemoteBackend` and `FullName` when creating/ updating `dbtcloud_repository`
- Added test at `dbt_cloud` client level to cover this case and see that the update request is sent accordingly
- Added fix to the repository update method that wasn't updating all the fields accordingly
- Removed `RequiresReplace()` for the `repository_id` field in the `dbtcloud_project_repositoryso` so that the whole resource doesn't get recreated for just updating the `pull_request_url_template` for example. (If a field from the `dbtcloud_repository` with `RequiresReplace()` will change, the `dbtcloud_project_repository` will change as well)

- [x] Manual testing
- [x] Added tests


gravelbit (approved):
👌

PR #466, opened 2025-06-18
SIGN-163 Create SL Credential for Postgres resource

teodoravochin-dbtlabs:
Created Postgres Semantic Layer Credential resource.
Created Acceptance tests for Postgres SL Credential.
Created docs.


pkg/framework/objects/postgres_credential/schema.go
```diff
@@ -68,17 +71,25 @@ var resourceSchema = resource_schema.Schema{
 			Description: "The system Postgres/Redshift/AlloyDB credential ID.",
```
vladmld, line 71: Can we leave here just "The system Postgres credential ID."? Or are the others true also?

vladmld (approved):

PR #468, opened 2025-06-20
SIGN-200 Fixed Missing Required Attribute Issue

teodoravochin-dbtlabs:
Customer received error message '`schema` must be provided when `semantic_layer_credential` is false.' even if they provided the value in .tfvars file

SemanticLayerCredentialValidator which was added on some fields was throwing error before reading the values assigned from .tfvars file.


victorasu (approved):
LGTM! 👍

victorasu:
Please also make sure the docs and examples match the implementation.

PR #484, opened 2025-07-09
Issue 478 Revert changing repository id to recreate the project repository

vladmld:
Reverted a past change that was not recreating the `project_repository` resource when recreating the `repository` resource.


victorasu (approved):
LGTM!

PR #502, opened 2025-07-25
Update environment.md - docs for credential_id

patkearns10:
The `credential_id` had documentation for `project_id`. Fixed that.

But, I am curious about something...

Here it says `credential_id`, but [via the api](https://docs.getdbt.com/dbt-cloud/api-v3#/operations/Retrieve%20Environment), it is `credentials_id`. 
I see both are used throughout the codebase. Which is correct?

https://github.com/search?q=repo%3Adbt-labs%2Fterraform-provider-dbtcloud+credential_id&type=code 
vs.
https://github.com/search?q=repo%3Adbt-labs%2Fterraform-provider-dbtcloud+credentials_id&type=code

### DIFF
--- a/.changes/unreleased/Documentation-20250804-104556.yaml
+++ b/.changes/unreleased/Documentation-20250804-104556.yaml
@@ -1,0 +1,3 @@
+kind: Documentation
+body: Doc updates
+time: 2025-08-04T10:45:56.790732+03:00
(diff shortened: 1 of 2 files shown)


adrianburusdbt:
Hey Pat, thanks for fixing the issue in the docs. I think if you just run `changie new` and fill in the change details it will pass the build too.
The `credential_id` naming is used for all credentials (postgres, snowflake, etc), and `credentials_id` is used for repository and environment. I don't have any context on why this is so, I don't think it makes any functional difference and is probably more a case of what the people working on them considered.
FWIW, the docs on each of the two types is consistent with the code :)

PR #504, opened 2025-07-29
fix credential ID bug in postgres and databricks Update functions

gravelbit:
Also add unit tests that validate expected dbt calls in the broken workflow scenario.


dbt_cloud/testutil/postgres_credential_bug_fix_test.go
adrianburusdbt, line 1: Typically the test files sit next to the resources and datasources. Did you leave them in testutil on purpose?

gravelbit: It was easy to built them that way then I forgot to move them. I now moved them to the right place with the latest commit. Thanks for the heads up.

adrianburusdbt (approved):

PR #517, opened 2025-08-26
Added data source to get all private link endpoints

adrianburusdbt:
Added data source to get all private link endpoints


victorasu (approved):
It could potentially be worth adding an example as well. :shipit:

PR #554, opened 2025-11-11
Add skip_credentials_validation flag to provider

amirgo1:
This change adds a new optional boolean flag `skip_credentials_validation` to the dbt Cloud provider, similar to the AWS provider implementation.

When set to true, the provider will skip the API call that validates credentials during initialization. This is useful for:
- Testing scenarios where credential validation is not needed
- Custom dbt Cloud API implementations without standard authentication
- Environments where the validation endpoint may not be available

The flag defaults to false to maintain backward compatibility.

Changes:
- Added skip_credentials_validation to provider schema
- Added SkipCredentialsValidation to provider model
- Updated Configure method to pass flag to client
- Modified NewClient signature to accept skipCredentialsValidation parameter
- Updated validation logic to respect the skip flag
- Updated provider documentation


adrianburusdbt:
Looks good, thanks!

PR #557, opened 2025-11-17
Add apache spark credentials

### LINKED ISSUE #497: Add support for Apache Spark credentials

I'm running into an issue when trying to manage Apache Spark credentials and environment-specific schemas through Terraform because the `dbtcloud_databricks_credential` resource doesn't support `adapter_type = "spark" (or apache_spark?)`.
Currently, when attempting to create Spark credentials:
```
resource "dbtcloud_databricks_credential" "apache_spark" {
  project_id   = dbtcloud_project.project.id
  token        = var.credentials.token
  schema       = var.credentials.schema
  adapter_type = "spark"
}
```
The provider fails with:

```
Error: Spark adapter not supported

Spark adapter is not supported currently for global connections credentials. 
Please raise a GitHub issue if you need it
```

This means Apache Spark deployment environments get credential_id = null and cannot have environment-specific schemas (staging/production) managed through Terraform, forcing manual configuration in the dbt Cloud UI.

The ability to set a token and a schema (on the deployment envs) to an apache spark project (any way possible)

## Additional context
Technical Context:
- Provider Version: dbt-labs/dbtcloud ~> 1.0
- Source Code: The validation appears to be in `pkg/framework/objects/databricks_credential/schema.go`
- Error Message: Explicitly suggests "Please raise a GitHub issue if you need it"

My Apache Spark Global Connection:
```
resource "dbtcloud_global_connection" "apache_spark_connection" {
  name = "My Apache Spark connection
  apache_spark = {
    method          = "http"
    host            = "my-spark-host.com"
    cluster         = "123-12345-example"
    connect_timeout = 100
  }
}
```

adrianburusdbt:
### Summary
Closes #497
This PR adds native support for Apache Spark credentials by introducing a new dbtcloud_spark_credential resource and data source. Previously, users had to manually configure Spark credentials in the dbt Cloud UI as the provider only supported Databricks credentials with the databricks adapter type.

### Motivation
Users running Apache Spark deployments were unable to manage credentials through Terraform because:
The dbtcloud_databricks_credential resource explicitly blocked adapter_type = "spark" with an error message
Apache Spark credentials require different adapter configuration (apache_spark_v0) than Databricks (databricks_v0)
Mixing both credential types in a single resource would create confusion and maintenance issues

### New Resources
**1. dbtcloud_spark_credential Resource**
- Dedicated resource for managing Apache Spark credentials
- Uses apache_spark_v0 adapter version automatically
- Supports full CRUD operations (Create, Read, Update, Delete)
- Simpler schema than Databricks (no catalog or auth_type fields)

**2. dbtcloud_spark_credential Data Source**
- Query existing Apache Spark credentials
- Returns credential details for use in other resources

### DIFF
--- a/pkg/dbt_cloud/spark_credential.go
+++ b/pkg/dbt_cloud/spark_credential.go
@@ -48,7 +48,7 @@ type SparkCredentialGlobConn struct {
 	CredentialDetails AdapterCredentialDetails `json:"credential_details"`
 }
 
-type SparkCredentialGLobConnPatch struct {
+type SparkCredentialGlobConnPatch struct {
 	ID                int                      `json:"id"`
 	CredentialDetails AdapterCredentialDetails `json:"credential_details"`
 }
@@ -149,7 +149,7 @@ func (c *Client) CreateSparkCredential(
 func (c *Client) UpdateSparkCredentialGlobConn(
 	projectId int,
 	credentialId int,
-	sparkCredential SparkCredentialGLobConnPatch,
+	sparkCredential SparkCredentialGlobConnPatch,
 ) (*SparkCredential, error) {
 	sparkCredentialData, err := json.Marshal(sparkCredential)
 	if err != nil {
--- a/pkg/framework/objects/spark_credential/resource.go
+++ b/pkg/framework/objects/spark_credential/resource.go
@@ -283,7 +283,7 @@ func (d *sparkCredentialResource) updateGlobal(ctx context.Context, plan, state 
 			}
 		}
 
-		sparkPatch := dbt_cloud.SparkCredentialGLobConnPatch{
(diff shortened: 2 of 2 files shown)


pkg/dbt_cloud/spark_credential.go
```diff
@@ -0,0 +1,263 @@
...
+		Threads:           NUM_THREADS_CREDENTIAL,
+		CredentialDetails: credentialDetails,
+	}
+
+	newSparkCredentialData, err := json.Marshal(newSparkCredential)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(
```
gravelbit: It will be nice to add `ctx context.Context` param to this method and use `NewRequestWithContext` instead of the simple one to have graceful request cancelation instead of relying on time out in some corner-cases.

Issue #553, opened 2025-11-07
Plugin crash and environment variable creation conflict on dbtcloud provider v1.3.0 (resource-not-found and SIGSEGV during apply)


ByteMeDirk:
## Describe the bug

Running terraform apply to create dbtcloud_environment and associated variables results in resource-not-found errors and
a plugin crash (SIGSEGV) when the environment variable already exists in the project/account (in most cases the
environment variable does not exist). The provider does not handle the duplicate resource gracefully and subsequently
crashes.

## Error message

Configure required providers as below:

```terraform
required_providers {
  dbtcloud = {
    source  = "dbt-labs/dbtcloud"
    version = "1.3.0"
  }
}
```

Attempt to create environment variables with names that already exist in our dbt Cloud project (i.e., DBT_ENVIRONMENT,
DBT_TABLE_EXPIRATION_HOURS). Note: these do not exist in our project yet.

```terraform
resource "dbtcloud_environment_variable" "dbt_environment" {
  name       = "DBT_ENVIRONMENT"
  project_id = "<project_id>"
  # etc.
}
```

Run terraform apply

## Resource configuration

```terraform
terraform {
  required_version = "~>1.13.4"

  required_providers {
    dbtcloud = {
      source  = "dbt-labs/dbtcloud"
      version = "1.3.0"
    }
  }
  # etc.
}

# Environment Variable: DBT_ENVIRONMENT

resource "dbtcloud_environment_variable" "dbt_environment" {
  environment_values = {
    Development              = "development"
    "Development Deployment" = "development"
    "Production Deployment"  = "production"
    project                  = "development"
  }
  name       = "DBT_ENVIRONMENT"
  project_id = dbtcloud_project.data_warehouse.id
}

# Environment Variable: DBT_TABLE_EXPIRATION_HOURS

resource "dbtcloud_environment_variable" "dbt_table_expiration_hours" {
  environment_values = {
    Development              = ""
    "Development Deployment" = "none"
    "Production Deployment"  = "none"
    project                  = "none"
  }
  name       = "DBT_TABLE_EXPIRATION_HOURS"
  project_id = dbtcloud_project.data_warehouse.id
}
```

## Expected behavior

Provider should return a useful error and fail gracefully or allow conditional updates/imports. It should not crash (
SIGSEGV).

Actual behavior:

- Error: environment variable already exists
- Error: Plugin did not respond

Crash log:

```text
[...]

Issue #534, opened 2025-09-17
Unable to create external email notifications on dbt Cloud terraform provider 1.2.2


sanjay-lendi:
## Describe the bug
Getting a plugin error when creating `dbtcloud_notification` using the latest terraform provider 1.2.2 with `notification_type = 4`

## Error message
Here is the buildkite logs showing the plugin error. 

```sh
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:

  + create

  ~ update in-place



Terraform will perform the following actions:



  # dbtcloud_job.dev_job_full_refresh will be updated in-place

  ~ resource "dbtcloud_job" "dev_job_full_refresh" {

        id                     = 111111

        name                   = "Manual Full Refresh - DEV"

      - schedule_days          = [

          - 0,

          - 1,

          - 2,

          - 3,

          - 4,

          - 5,

          - 6,

        ] -> null

      ~ schedule_type          = "days_of_week" -> "every_day"

      ~ self_deferring         = false -> (known after apply)

        # (19 unchanged attributes hidden)

    }



  # dbtcloud_notification.dev_jobs_email_notification will be created

  + resource "dbtcloud_notification" "dev_jobs_email_notification" {

      + external_email    = "channel@slack.com"

      + id                = (known after apply)

      + notification_type = 4

      + on_cancel         = [

          + 123456,

          + 123457,

        ]

      + on_failure        = [

          + 123456,

          + 123457,

        ]

      + on_success        = []

      + on_warning        = []

      + state             = 1

      + user_id           = 3333

    }



  # dbtcloud_notification.production_jobs_email_notification will be created

  + resource "dbtcloud_notification" "production_jobs_email_notification" {

      + external_email    = "channel@slack.com"

      + id                = (known after apply)

      + notification_type = 4

      + on_cancel         = [

          + 223456,

          + 223457,

        ]

      + on_failure        = [

          + 223456,

          + 223457,

        ]

      + on_success        = []

      + on_warning        = []

      + state             = 1

      + user_id           = 3333

    }



  # dbtcloud_notification.staging_jobs_email_notification will be created

  + resource "dbtcloud_notification" "staging_jobs_email_notification" {

      + external_email    = "channel@slack.com"

      + id                = (known after apply)

      + notification_type = 4

      + on_cancel         = [

          + 33348,

          + 33349,

        ]

      + on_failure        = [

          + 33348,

          + 33349,

        ]

[...]

Issue #522, opened 2025-09-05
[TF] Plugin failure with job creation resources


CrisNavasM:
## Describe the bug
It fails when applying a plan containing job resources creation (plus, it actually creates them half way in dbt cloud but no stated of that is saved, assuming due to the failure)

## Error message
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9f042c]
> Error: The terraform-provider-dbtcloud_v1.2.1 plugin crashed!

## Resource configuration
```hcl
# Daily prod build
resource "dbtcloud_job" "daily_prd" {
  for_each       = var.projects
  name           = "${each.key} - daily prd"
  project_id     = local.project_ids[each.key]
  environment_id = local.env_ids["${each.key}-prd"]
  execute_steps = [
    "dbt build" #TODO: Add `--select [dp-name]`
  ]
  triggers = {
    schedule             = true
    on_merge             = false
    git_provider_webhook = false
  }
  schedule_days  = [0, 1, 2, 3, 4, 5, 6]
  schedule_type  = "days_of_week"
  schedule_hours = [0]
}

# Merge into prod
resource "dbtcloud_job" "merge_into_prd" {
  for_each                 = var.projects
  name                     = "${each.key} - merge into prd"
  project_id               = local.project_ids[each.key]
  environment_id           = local.env_ids["${each.key}-prd"]
  deferring_environment_id = local.env_ids["${each.key}-prd"]
  run_compare_changes      = true
  generate_docs            = true
  execute_steps = [
    "dbt build --select state:modified+ --fail-fast"
  ]
  triggers = {
    schedule             = false
    on_merge             = true
    git_provider_webhook = false
  }
}

# Ad-hoc prod
resource "dbtcloud_job" "adhoc_prd" {
  for_each       = var.projects
  name           = "${each.key} - ad-hoc prd"
  project_id     = local.project_ids[each.key]
  environment_id = local.env_ids["${each.key}-prd"]
  execute_steps = [
    "dbt build" #TODO: Modify as required
  ]
  triggers = {
    schedule             = false
    on_merge             = false
    git_provider_webhook = false
  }
}

# MR job in UAT on PR open/update
resource "dbtcloud_job" "ci_uat" {
  for_each                 = var.projects
  name                     = "${each.key} - MR job (uat)"
  project_id               = local.project_ids[each.key]
[...]

PR #564, opened 2025-11-26
Null id handling mechanism

adrianburusdbt:
### Add API Response Validation to Prevent Nil Pointer Panics
**Summary**
Implements API client-level validation to catch nil pointer dereferences before they cause provider crashes. This centralized approach validates required fields in API responses and provides detailed, actionable error messages to users.
**Issues Fixed**

- #534 
- #522 
- #553 

**Changes**
Added pkg/dbt_cloud/null_id_validation.go - Validation framework with custom error types and response validators
Added pkg/dbt_cloud/null_id_validation_test.go - Comprehensive test coverage (9 test functions, 35+ test cases, 100% passing)
Updated notification.go - Added validation for Create/Read/Update operations
Updated model_notifications.go - Added validation for Create/Update operations
Updated project.go - Added validation for Create/Read/Update operations

### DIFF
--- a/pkg/dbt_cloud/notification.go
+++ b/pkg/dbt_cloud/notification.go
@@ -155,9 +155,13 @@ func (c *Client) UpdateNotification(
 		return nil, err
 	}
 
-	// Validate the response has required fields
-	if err := ValidateNotificationResponse(notificationResponse.Data.Id, &notificationResponse.Data); err != nil {
-		return nil, err
+	// Skip validation for delete operations (STATE_DELETED) as the API may return nil ID
+	// and we don't use the response data for deleted resources
+	if notification.State != STATE_DELETED {
+		// Validate the response has required fields
+		if err := ValidateNotificationResponse(notificationResponse.Data.Id, &notificationResponse.Data); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/dbt_cloud/null_id_validation.go
```diff
@@ -0,0 +1,314 @@
...
+			Response:     response,
+		}
+	}
+
+	return nil
+}
+
+// ValidateModelNotificationsResponse validates that a model notifications response has required fields
+// It accepts the ID pointer and the full response for error reporting
+func ValidateModelNotificationsResponse(id *int, response interface{}) error {
```
tomighita: Shouldn't we collapse all these functions into a more generic function which just checks that a response is not null and that the id is present? Do we need a function for each response type which does the same thing?

adrianburusdbt:
> Left a few comments about the amount of custom functions.
> 
> Also, should we consider some proper validation frameworks for our objects [eg [this](https://github.com/go-playground/validator)]? I feel like it would be very easy to extend in the future.

That's a good idea, let me try it out and come back with a conclusion.

Issue #571, opened 2025-12-02
Plan-time validation of dbtcloud_job execute_steps


ginone:
## Is your feature request related to a problem? Please describe.

We manage our dbt Cloud account via the Terraform dbt Cloud provider (used through CDK for Terraform). Our merge request CI runs:

- `cdktf synth` (generate Terraform)
- `cdktf diff` (plan) to gate merges to the main branch.

### Issue:

- Certain invalid job steps are rejected by the dbt Cloud UI but are not caught during synth/diff. For example: `execute_steps = [ "dbt --version" ]` The UI correctly returns “One of the commands is not a valid dbt command,” but both `cdktf synth` and `cdktf diff` succeed. The error only appears during `cdktf deploy`, i.e., after merging to main.

### Steps to reproduce:

1. Define a `dbtcloud_job` with `execute_steps = ["dbt --version"]`.
2. Run `cdktf synth` and `cdktf diff` — both pass without errors.
3. Run `cdktf deploy` — the apply fails with the validation error from dbt Cloud.

## Describe the solution you'd like
Expected behavior:

- Invalid job commands should be surfaced during plan/diff (or earlier) so CI can block the merge.

Actual behavior:

- No error until apply/deploy.

Impact:

- Broken job definitions can be merged to main before being caught, causing failed deployments.

Request:

- Could the provider perform plan-time validation of execute_steps to mirror the UI’s command validation?
- Alternatively, could you expose a public validation endpoint that the provider (or CI) could call pre-apply?

## Describe alternatives you've considered
In our merge request CI, we could add a step that would dynamically create fake “test” jobs in dbt Cloud with the `execute_steps` from actual job definitions and then remove them. That is quite a lot of overhead though and it would be best if there were a way to validate `execute_steps` without creating anything in dbt Cloud.

victorasu:
Hey @ginone, sorry for the late reply. The plan-time validation was added as an optional feature on the provider that can be toggled per-resource. The docs were changed to reflect the added defensive capabilities, you can read more about the changes in #572.

We'll be releasing a new version of the provider this week, so this capability will be available promptly. Thank you for your contribution!

PR #572, opened 2025-12-03
#571 Job execute_steps validation

victorasu:
These changes conform with dbt-cloud's [implementation](https://github.com/dbt-labs/dbt-cloud/blob/master/sinter/clients/dbt.py#L132) for validating execution steps on Jobs.


b-per:
I am personally not convinced that we should by default add such validations (even if there was a feature request about this) 😄

This would mean that if in the future we add new commands/flags to dbt then:
1. we would need to do a new release of the terraform provider with those new commands/flags
2. people will have to update to the latest provider version (with potential breaking changes) in order to use the new commands/flags

If we really want to add it I would actually make it an "opt-in" behavior that is off by default.

PR #579, opened 2025-12-09
Delete lingering Group if permission update fails

victorasu:
In the case where a permission cannot be added to a certain group, for reasons like a missing account-level setting, feature flag and so on, the Group gets created but when the Permission addition fails the Group remains empty and is not added to the TF state.
This change aims to solve that issue by deleting the lingering Group in case of permission update failure.


tudor-scurtu (approved):
Thank you! :shipit:

PR #587, opened 2025-12-15
Use state for job_type since it requires replace

adrianburusdbt:
**Problem:**
Upgrading from provider v1.4.2 to v1.5.0 caused all existing jobs to be marked for recreation due to `job_type` changing from its stored value (e.g., "`other`") to (known after apply).

**Root Cause:**
The `job_type` attribute is Optional + Computed with `RequiresReplace()`, but was missing `UseStateForUnknown()`. When the new `validate_execute_steps` attribute was added in v1.5.0, it triggered Terraform to re-evaluate the plan, causing `job_type` to become "`unknown`" instead of preserving the state value.

**Fix:**
Added `stringplanmodifier.UseStateForUnknown()` to the job_type attribute, preserving the existing state value during planning.
The underlying issue was that we were replacing the job every time 'something' changed. In this case, it was adding the validation_execution_steps value, but it surfaced some gaps in the update logic as well.
The final fix involved matching the job_type change rules, so that they are the same as in dbt cloud.

### DIFF
--- a/pkg/framework/objects/job/resource.go
+++ b/pkg/framework/objects/job/resource.go
@@ -144,16 +144,20 @@ func (j *jobResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanReq
 	// Validate job_type field changes if the field is being explicitly set
 	// Note: If plan.JobType is set but state.JobType is null (first time setting it),
 	// the validation will happen in Update against the actual server value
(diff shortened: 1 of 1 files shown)


victorasu (approved):
I'm flabbergasted. :shipit:

PR #591, opened 2025-12-15
Ensure use_custom_branch and custom_branch value are cohesive

adrianburusdbt:
**Problem**
When `custom_branch` was set but `use_custom_branch` was false (or not set), the environment configuration would be inconsistent - the API would receive conflicting values, causing confusing behavior in the dbt Cloud UI.

**Solution**
Added schema validators to ensure `use_custom_branch` and `custom_branch` are used consistently:
Error if `custom_branch` is set but `use_custom_branch` is false
Error if `use_custom_branch` is true but `custom_branch` is not set

**Changes**
Added validators/custom_branch_validator.go with CustomBranchValidator and UseCustomBranchValidator
Added validators to the use_custom_branch and custom_branch schema fields
Added acceptance tests for both validation scenarios
Fixed existing test that was setting custom_branch without use_custom_branch = true

### DIFF
--- a/.changes/unreleased/Behind the scenes-20251215-172613.yaml
+++ b/.changes/unreleased/Behind the scenes-20251215-172613.yaml
@@ -1,0 +1,3 @@
+kind: Behind the scenes
+body: Custom branch and use custom branch must match
+time: 2025-12-15T17:26:13.609056+02:00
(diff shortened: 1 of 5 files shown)


gravelbit (approved):
👍

PR #598, opened 2025-12-19
Treat unknown custom_branch

victorasu:
**Problem**:
When `custom_branch` was being supplied as a variable, the provider would throw a validation error stating the value must be specified, even if the user provided a valid value.
```hcl
resource "dbtcloud_environment" "production" {
  name            = "Production"
  ...
  use_custom_branch = true
  custom_branch     = var.custom_branch
}
``` 
```
│ Error: Missing custom_branch
│ 
│   with module.data_engineering_foundational_project.dbtcloud_environment.production,
│   on modules/dbt_project/main.tf line 133, in resource "dbtcloud_environment" "production":
│  133:   use_custom_branch = true
│ 
│ When use_custom_branch is set to true, custom_branch must be specified.
```

**Root cause**:
When using `custom_branch` as a variable through a module, it would have an Unknown state at validation through CustomBranchValidator. This state was not being verified in the validator and lead to the validator evaluating it as non-existent, even if the value was being supplied. This caused `customBranch.ValueString() != ""` to become true, as an empty string is the default value for String Unknowns, therefore asserting that `custom_branch` was not provided.

**Fix**:
Add a check specifically asserting whether `custom_branch` is known yet or not. If it is not, disregard current validation attempt and reevaluate once the value becomes known.


victorasu:
@adrianburusdbt it was reported in Slack. I've shared a link privately.

PR #601, opened 2026-01-12
Fix: BigQuery credentials compatible with use_latest_adapter = true

adrianburusdbt:
**Problem**
BigQuery credentials created via Terraform were incompatible with global connections using use_latest_adapter = true, causing jobs to fail.

**Solution**
Added connection_id attribute to dbtcloud_bigquery_credential. When provided, the provider auto-detects the adapter version and creates credentials with the correct format.

**Usage**
```
resource "dbtcloud_bigquery_credential" "cred" {  
  project_id    = dbtcloud_project.project.id  
  dataset       = "my_dataset"  
  num_threads   = 12  
  connection_id = dbtcloud_global_connection.bq.id  # Enables auto-detection
}
```

**Changes**

Added `connection_id` attribute to `dbtcloud_bigquery_credential`
Auto-detect adapter version from global connection
Support both legacy (v0) and new (v1) credential formats
Added acceptance tests

### DIFF
--- a/pkg/framework/objects/partial_notification/resource_acceptance_test.go
+++ b/pkg/framework/objects/partial_notification/resource_acceptance_test.go
@@ -2,11 +2,12 @@ package partial_notification_test
 
 import (
 	"fmt"
-	"github.com/dbt-labs/terraform-provider-dbtcloud/pkg/framework/acctest_config"
 	"regexp"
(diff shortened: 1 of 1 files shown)


victorasu (approved):
With limited manual testing looks good! 🚀

PR #604, opened 2026-01-14
Handle resource removal from state if resource not found

adrianburusdbt:
**Summary**
Adds proper resource-not-found error handling to dbtcloud_semantic_layer_credential_service_token_mapping resource's Read() function, consistent with other semantic layer resources fixed in #597.

**Changes**
When the API returns "resource-not-found", the resource is now gracefully removed from state with a warning instead of failing with an error
This aligns with standard Terraform patterns for handling resources deleted out-of-band


victorasu (approved):
I think we should have this handling on all other resources too.

PR #605, opened 2026-01-16
Enable execution block and timeout_seconds

### LINKED ISSUE #596: Job timeout_seconds appears deprecated but has no alternative
## Describe the bug
The documentation and validation of the [`timeout_seconds`](https://registry.terraform.io/providers/dbt-labs/dbtcloud/latest/docs/resources/job#timeout_seconds-1) job property indicates it is deprecated, but there appears to be no alternative way or location to set this. Tried looking in the code to check if it was just missing documentation but couldn't find an alternative implementation.

## Error message
`"timeout_seconds" is deprecated: Reason: "[Deprectated - Moved to execution.timeout_seconds] Number of seconds to allow the job to run before timing out"`

## Resource configuration
```hcl
resource "dbtcloud_job" "this" {
  ...
  timeout_seconds = 3600
}
```

## Expected behavior
Either no deprecation warning, or a new location where to define this

`provider[registry.terraform.io/dbt-labs/dbtcloud] ~> 1.5.0`

## Additional context
N/A

adrianburusdbt:
This PR fixes the issue where `timeout_seconds` was marked as deprecated with "Moved to execution.timeout_seconds" but the execution block was never implemented.

**Changes**:
- Added execution block with timeout_seconds attribute to the job resource
- Marked top-level timeout_seconds as deprecated with proper migration guidance
- Full backward compatibility: existing configs using timeout_seconds continue to work
- When both are set, execution.timeout_seconds takes precedence

Closes #596

### DIFF
--- a/.changes/unreleased/Changes-20260116-155106.yaml
+++ b/.changes/unreleased/Changes-20260116-155106.yaml
@@ -1,0 +1,3 @@
+kind: Changes
+body: Fixed timout_seconds missing field
+time: 2026-01-16T15:51:06.20357+02:00
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -45,6 +45,11 @@ resource "dbtcloud_job" "daily_job" {
   schedule_days  = [0, 1, 2, 3, 4, 5, 6]
   schedule_type  = "days_of_week"
   schedule_hours = [0]
+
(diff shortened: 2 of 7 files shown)


vladmld (approved):
👍🏼

Issue #585, opened 2025-12-11
Add support for Catalog Metadata Ingestion


joaoclemencio-dbtlabs:
## Is your feature request related to a problem? Please describe.
[External Metadata Ingestion](https://docs.getdbt.com/docs/explore/external-metadata-ingestion#prerequisites) isn't supported yet by the [`dbtcloud_account_features` resource](https://registry.terraform.io/providers/dbt-labs/dbtcloud/latest/docs/resources/account_features).

## Describe the solution you'd like
Not sure if this should all be in a single issue, but as I can find there are at least 3 configurations required:
- Add the flag to the list of supported features. In the UI, this is triggered by the same API call as the others (POST to `https://{dbtcloud_host}/api/private/accounts/{account_id}/features/`) with the payload `{feature: "catalog-ingestion", value: true}`
- Add a database credentials resource to query the database metadata. In the UI, this is set in the Connection, specifically via a PATCH to `https://{dbtcloud_host}/api/v3/accounts/{account_id}/platform-metadata-credentials/`
- Add a `catalog-configs` resource that applies filters to the connection the credentials have been applied to. This is also a PATCH, but to `https://{dbtcloud_host}/api/v3/accounts/{account_id}/connections/{connection with metadata credentials}/catalog-configs/` with an allow/deny list payload:
```
{
  "database_allow": [],
  "database_deny": [],
  "schema_allow": [],
  "schema_deny": [],
  "table_allow": [],
  "table_deny": [],
  "view_allow": [],
  "view_deny": []
}
```

## Describe alternatives you've considered
Manually setting it in the UI; potentially querying the API separately but would prefer terraform

PR #606, opened 2026-01-19
Added support for metadata ingestion

adrianburusdbt:
Addressing everything in #585:
Add the flag to the list of supported features. In the UI, this is triggered by the same API call as the others (POST to https://{dbtcloud_host}/api/private/accounts/{account_id}/features/) with the payload {feature: "catalog-ingestion", value: true}
Add a database credentials resource to query the database metadata. In the UI, this is set in the Connection, specifically via a PATCH to https://{dbtcloud_host}/api/v3/accounts/{account_id}/platform-metadata-credentials/
Add a catalog-configs resource that applies filters to the connection the credentials have been applied to. This is also a PATCH, but to https://{dbtcloud_host}/api/v3/accounts/{account_id}/connections/{connection with metadata credentials}/catalog-configs/ with an allow/deny list payload:

### DIFF
--- a/pkg/framework/objects/ip_restrictions_rule/resource_acceptance_test.go
+++ b/pkg/framework/objects/ip_restrictions_rule/resource_acceptance_test.go
@@ -11,6 +11,8 @@ import (
 )
 
 func TestAccDbtCloudIPRestrictionsRuleResource(t *testing.T) {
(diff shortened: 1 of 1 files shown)


gravelbit (approved):
Great implementation 🙏. I left some questions/ideas inline to follow in future updates.

Issue #610, opened 2026-01-21
Permission Deduplication Missing in Create Operations for Partial Permission Resources


trouze:
## Summary

The `dbtcloud_scim_group_partial_permissions` and `dbtcloud_group_partial_permissions` resources can create duplicate permission records when multiple Terraform resources manage the same group concurrently, or when permissions are created/destroyed repeatedly. This can lead to exponential permission growth (e.g., 50+ identical permissions accumulating over time), eventually causing API timeouts and operational issues.

## Affected Resources

- `dbtcloud_scim_group_partial_permissions`
- `dbtcloud_group_partial_permissions`

## Root Cause

The `Create` operation in both resources uses Go's `append()` to combine remote permissions with new permissions:

```go
// BUGGY CODE
allPermissions := append(remotePermissions, missingPermissions...)
```

The `append()` function does not deduplicate, so if the same permission already exists in `remotePermissions` (which can happen due to timing or previous operations), it gets added again. Over many Terraform operations, this leads to exponential growth.

### How Permissions Are Duplicated

The issue occurs when:

1. **Resource A** reads the group, sees permissions `[X]`, adds permission `Y` → API now has `[X, Y]`
2. **Resource B** reads the group (during the same apply or later), sees `[X, Y]`, tries to add `Y` again
3. Due to timing or comparison logic, `Y` is considered "missing" and appended
4. API now has `[X, Y, Y]` ← **duplicate created**
5. This compounds over multiple apply cycles

This is particularly problematic when:
- Multiple `*_partial_permissions` resources manage the same group (federated pattern)
- Terraform apply operations run concurrently or in quick succession
- Resources are destroyed and recreated frequently

## Permission Uniqueness Definition

A permission is considered unique based on **all four** of these attributes:

1. **`permission_set`** - The role type (e.g., `account_admin`, `developer`, `analyst`)
2. **`project_id`** - The specific project ID (or `null` for account-level)
3. **`all_projects`** - Boolean flag for account-wide permissions
4. **`writable_environment_categories`** - Array of environment types (e.g., `["development", "staging"]`)

Two permissions are duplicates **only if all four attributes match exactly**. This ensures we never accidentally deduplicate truly distinct permissions like:
- `developer` on project 123 vs project 456
- `developer` with `["development"]` vs `["development", "staging"]`
- `developer` on specific project vs all projects

### Why Not Track Permission IDs?

[...]

PR #611, opened 2026-01-21
Fix permission duplication in partial_permissions resources

trouze:
Proposed resolution to #610 

- Replace append() with UnionBy() in Create to prevent new duplicates
- Add UniqBy() deduplication after reading remote permissions for self-healing
- Add UniqBy() helper function to deduplicate slices with custom comparison
- Add unit tests for deduplication logic
- Add changie entry

Fixes issue where permissions could duplicate exponentially when multiple resources manage the same group or during repeated create/destroy cycles. The fix is self-healing and will automatically clean up existing duplicates.

### DIFF
--- a/pkg/framework/objects/group_partial_permissions/resource.go
+++ b/pkg/framework/objects/group_partial_permissions/resource.go
@@ -106,6 +106,9 @@ func (r *groupPartialPermissionsResource) Read(
 
 	// we set the "partial" values
 	remotePermissions := group.ConvertGroupPermissionDataToModel(retrievedGroup.Permissions)
+	
+	// Deduplicate remote permissions in case they were created by a buggy version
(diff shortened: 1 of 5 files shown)


pkg/framework/objects/group_partial_permissions/resource.go
```diff
@@ -371,6 +387,9 @@ func (r *groupPartialPermissionsResource) Update(
 	planPermissions := plan.GroupPermissions
 
 	remotePermissions := group.ConvertGroupPermissionDataToModel(retrievedGroup.Permissions)
+	
+	// Deduplicate remote permissions in case they were created by a buggy version
+	remotePermissions = helper.UniqBy(remotePermissions, group.CompareGroupPermissions)
```
adrianburusdbt: Is there no scenario in which there could be 'duplicate' permissions and removing them can break existing logic / behaviors?

PR #619, opened 2026-01-30
Updating accountID from int to int64

joeguaneri:

PR #622, opened 2026-02-09
Replace all AccountID ints with int64

adrianburusdbt:
Builds on #619 by @joeguaneri. Thanks for the initial work!

This PR includes their changes plus the requested follow-ups (remaining AccountID conversions and determineInt64Value in acctest_config).


vladmld (approved):
Looks good! 👍🏼

PR #667, opened 2026-04-09
fix: remove connection_id default of 0 in dbtcloud_environment (closes #665)

### LINKED ISSUE #665: fix: remove connection_id default of 0 in dbtcloud_environment to prevent inconsistent result after apply
## Summary

`dbtcloud_environment` currently has `Default: int64default.StaticInt64(0)` on `connection_id`. This causes Terraform to report an "inconsistent result after apply" error when the API auto-assigns a connection to the environment (returns a non-zero value) but Terraform's plan expected 0.

## Problem

When a user creates an environment without explicitly setting `connection_id`, the provider sends `0` to the API. The API may auto-assign a connection and return a non-zero `connection_id` in the response. Because the Terraform plan expected `0` (from the schema default) but the applied state contains the real ID, Terraform raises:

```
```

## Fix

- Remove `Default: int64default.StaticInt64(0)` from the `connection_id` attribute
- In `Create`, only send `connection_id` to the API when the plan value is non-null and non-unknown; otherwise send `0`
- After create/read, always store the API's returned value (or null if nil) rather than defaulting to 0
- Update `Update` to use null-safe comparison instead of direct value comparison for `connection_id` changes

## Files

- `pkg/framework/objects/environment/schema.go` — remove int64 default
- `pkg/framework/objects/environment/resource.go` — null-safe create/update handling

## Source

Developed on branch `fix/provider-bug-wsargent`.

trouze:
## Summary
- Removes `Default: int64default.StaticInt64(0)` from `connection_id` in `dbtcloud_environment` schema
- Adds `connectionIDFromAPI(*int) types.Int64` helper that maps both `nil` and `0` from the API to `types.Int64Null()` in Terraform state
- Fixes Create, Update, and Read methods to use null-safe handling for `connection_id`

Prevents the "inconsistent result after apply" error that occurs when the dbt Cloud API auto-assigns a non-zero connection ID after create.

Closes #665

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.changes/unreleased/Fixes-20260409-195804.yaml
+++ b/.changes/unreleased/Fixes-20260409-195804.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: Remove `connection_id` default of `0` in `dbtcloud_environment` to prevent "inconsistent result after apply" when the API auto-assigns a connection
+time: 2026-04-09T19:58:04.000000+00:00
--- a/pkg/framework/objects/environment/resource.go
+++ b/pkg/framework/objects/environment/resource.go
@@ -19,6 +19,15 @@ var (
 	_ resource.ResourceWithImportState = &environmentResource{}
 )
 
+// connectionIDFromAPI converts the API connection_id pointer to a types.Int64 value,
+// treating nil and 0 as "not set" (null) so Terraform doesn't see a spurious diff.
+func connectionIDFromAPI(id *int) types.Int64 {
+	if id != nil && *id != 0 {
+		return types.Int64Value(int64(*id))
+	}
+	return types.Int64Null()
+}
+
 func EnvironmentResource() resource.Resource {
 	return &environmentResource{}
 }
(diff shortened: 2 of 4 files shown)


will-sargent-dbtlabs:
Local verification — passes against a live dbt Cloud account.

**Setup:** Checked out `fix/issue-665-connection-id-default` at `72f0147` into a worktree and ran the targeted acceptance test against dbt Cloud account `dbt Labs - Sargent Sandbox` (`go1.26.3 darwin/arm64`).

**Fix confirmed in branch:**
- `int64default.StaticInt64(0)` is removed from `connection_id` in `pkg/framework/objects/environment/schema.go`.
- `connectionIDFromAPI` helper used in 3 places in `pkg/framework/objects/environment/resource.go` (Create, Update, Read) to map both `nil` and `0` from the API to `types.Int64Null()`.

**Test result:**
```
TF_ACC=1 go test -timeout 15m -count=1 -v -run '^TestAccDbtCloudEnvironmentResourceNoConnection$' ./pkg/framework/objects/environment/

=== RUN   TestAccDbtCloudEnvironmentResourceNoConnection
--- PASS: TestAccDbtCloudEnvironmentResourceNoConnection (16.04s)
PASS
ok      .../pkg/framework/objects/environment  16.838s
```

The "inconsistent result after apply" scenario the PR targets did not reproduce — Terraform now treats the API-returned non-zero `connection_id` as null in state when the user didn't set one, and no spurious drift on subsequent plan.

**One reviewer note (not a blocker for the fix itself):** this branch is stale relative to `main` — the diff also deletes `pkg/framework/objects/notification_setting/`, `.changes/v1.10.{1,2,3,4}.md`, and a release workflow change. Recommend a rebase onto current `main` before merge so the noise drops out.

/cc @adrianburusdbt

adrianburusdbt (approved):

PR #675, opened 2026-04-30
feat(environment_variable): detect changes to env vars in remote

trouze:
## Problem

The `dbtcloud_environment_variable` resource's `Read` function was calling the dbt Cloud API but only using the response to refresh the resource `id` — it silently discarded the returned `environment_values`. This meant Terraform state always reflected the last-applied config values, making drift detection impossible. Changes made outside of Terraform (e.g. via the dbt Cloud UI) would never surface in a subsequent `terraform plan`.

### Drift detection for non-secret env vars

`Read` now populates `state.EnvironmentValues` from the API response. On the next `terraform plan`, Terraform compares the refreshed state against config and surfaces any out-of-band changes as drift.

### Structural drift detection for secret env vars (`DBT_ENV_SECRET`)

The dbt Cloud API masks values for secret env vars, so raw value sync isn't possible. However, the API does return the key structure (which environments the var is defined for). `Read` now reconciles the API key set against the preserved state values:

- Environments removed from dbt Cloud → key dropped from state → drift surfaced on next plan
- Environments added in dbt Cloud → key added to state with empty string → drift surfaced on next plan
- Value changed in dbt Cloud → undetectable (fundamental API limitation — masked values can't be compared)

### Drift detection coverage

| Change type | Non-secret | Secret |
|---|---|---|
| Value changed outside Terraform | ✅ detected | ❌ API masks value |
| Environment key removed | ✅ detected | ✅ detected (key gone from API) |
| Environment key added | ✅ detected | ✅ detected (empty string surfaced) |
| Config value changed | ✅ detected | ✅ detected (state vs config diff) |

### Tests

- `TestAccDbtCloudEnvironmentVariableResourceDrift` — creates a non-secret env var, changes its values out-of-band via the API using `PreConfig`, then asserts `plancheck.ExpectNonEmptyPlan()` before the apply corrects the drift
- `getImportTestStep` — removed `environment_values` from `ImportStateVerifyIgnore` since non-secret values now round-trip correctly through import
- `getImportTestStepSecret` — new helper that retains the ignore for secret vars, used by `TestAccDbtCloudEnvironmentVariableResourceSecret`

### DIFF
--- a/pkg/framework/objects/environment_variable/resource.go
+++ b/pkg/framework/objects/environment_variable/resource.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/dbt-labs/terraform-provider-dbtcloud/pkg/dbt_cloud"
 	"github.com/dbt-labs/terraform-provider-dbtcloud/pkg/helper"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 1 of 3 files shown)


pkg/framework/objects/environment_variable/resource.go
```diff
@@ -157,6 +158,46 @@ func (r *environmentVariableResource) Read(
 	// Refresh state values
 	state.ID = types.StringValue(fmt.Sprintf("%d:%s", projectID, envVar.Name))
 
+	isSecret := strings.HasPrefix(envVar.Name, "DBT_ENV_SECRET_") ||
```
gravelbit: `dbt-cloud` uses `DBT_ENV_SECRET` as prefix for detecting secrets. It is better to drop the extra `_` when checking, for backend parity. Same for the other prefix.

gravelbit (approved):
Thanks for the fix. Please address my inline comments before merging.

pkg/framework/objects/environment_variable/resource.go
trouze: Good catch! I (not AI 😂) actually dreamt up that DBT_SECRET was a thing, so I removed that too.

PR #694, opened 2026-05-25
ZEI: Webhooks - Add archived webhook resurrection behavior to Webhook resource

victorasu:
Problem:
A new behavior that resurrects previously archived webhook subscriptions when `webhook_url` gets changed is being introduced in [dbt-cloud](https://github.com/dbt-labs/dbt-cloud/pull/17661). This behavior results in the provider seeing unexpected changes in the `is_active` field.

Solution:
Mark active as unknown if the `ClientURL` field is modified, for the field to consider the API response as the new source of truth. Also added an error message so users aren't taken by surprise by the change.

### DIFF
--- a/.changes/unreleased/Changes-20260525-164222.yaml
+++ b/.changes/unreleased/Changes-20260525-164222.yaml
@@ -1,3 +1,3 @@
 kind: Changes
-body: when updating clienturl on an inactive webhook, expect input from api
+body: When `client_url` changes on an inactive webhook, accept the reactivated state returned by the dbt Cloud API.
 time: 2026-05-25T16:42:22.059809+03:00
--- a/.changes/unreleased/Changes-20260525-164222.yaml
+++ b/.changes/unreleased/Changes-20260525-164222.yaml
@@ -1,3 +1,3 @@
 kind: Changes
-body: When `client_url` changes on an inactive webhook, accept the reactivated state returned by the dbt Cloud API.
+body: Error at plan time when `client_url` changes on an inactive webhook; users must set `active = true` to acknowledge that the dbt Cloud API will reactivate it.
 time: 2026-05-25T16:42:22.059809+03:00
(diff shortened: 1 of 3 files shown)


.changes/unreleased/Changes-20260525-164222.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Changes
+body: When `client_url` changes on an inactive webhook, accept the reactivated state returned by the dbt Cloud API.
```
vladmld: Would this be more explicit of the actual behavior? 🤔 
> Error at plan time when client_url changes on an inactive webhook; users must set active = true to acknowledge that the dbt Cloud API will reactivate it.

pkg/framework/objects/webhook/resource_acceptance_test.go
```diff
@@ -92,6 +92,34 @@ var modifyConfigTestStep = resource.TestStep{
 	),
 }
 
+// resurrectionPlanErrorTestStep verifies that changing client_url on an
+// inactive webhook while keeping active=false fails at plan time with a
+// clear error, because the dbt Cloud API would reactivate the webhook and
+// produce an inconsistent result.
+var resurrectionPlanErrorTestStep = resource.TestStep{
```
vladmld, line 99: **(Nit)** Enhance testing if you think it is worth it:

> Editing other fields on an inactive webhook. User wants to update description or event_types while keeping the webhook archived. Should succeed without forcing them to set active = true.

vladmld (approved):
Approved with comments. 👍🏼

pkg/framework/objects/webhook/resource_acceptance_test.go
victorasu: Not sure if that's worth it tbf, the backend check specifically only looks at the URL and doesn't touch anything else. I don't expect that to change in the future since there was a specific reason why URL was chosen in particular (i.e. because no other fields affect the webhook's interaction with the customer's server).

PR #696, opened 2026-05-26
fix(client): actually retry transient 5xx and 429 responses (#695)

### LINKED ISSUE #695: API client retry never fires on transient 5xx / 429 (control-flow bug in doRequestWithRetry)
## Summary

`pkg/dbt_cloud/client.go::doRequestWithRetry` advertises a retry-with-exponential-backoff policy via `MaxRetries`, `RetryIntervalSeconds`, and `RetriableStatusCodes` (provider config: `max_retries`, `retry_interval_seconds`, `retriable_status_codes`). In practice **the retry branch is unreachable** for any HTTP failure: the per-status `if res.StatusCode == … { return … }` blocks short-circuit before the retry decision runs. Every transient 5xx and 429 from `cloud.getdbt.com` is therefore surfaced as a one-shot Terraform diagnostic on the very first attempt.

This shows up most visibly as post-merge / scheduled CI flakes, but it also affects every customer whose `terraform apply` or `terraform plan` happens to coincide with a dbt Cloud API blip.

## Evidence

Two recent post-merge `unit (test-acceptance)` failures on `main`, no shared test, no shared code path — only shared symptom (a transient 5xx surfaced verbatim from the dbt Cloud API):

1. **2026-05-26**, run [26459657437](https://github.com/dbt-labs/terraform-provider-dbtcloud/actions/runs/26459657437), test `TestAccDbtCloudPartialNotificationResource` (`pkg/framework/objects/partial_notification`):
   ```
   Error: Error Reading dbt Cloud Project
     ...
     Could not read project ID 543791: unexpected status code 502: <html>
     <head><title>502 Bad Gateway</title></head>
     ...
     , URL:
     https://cloud.getdbt.com/api/v3/accounts/***/projects/543791/?include_related=[freshness_job_id,docs_job_id]
   ```

2. **2026-04-26** (scheduled), run [24947744549](https://github.com/dbt-labs/terraform-provider-dbtcloud/actions/runs/24947744549), tests `TestAccDbtCloudOpenAIIntegrationResource_OpenAI` and `TestAccDbtCloudOpenAIIntegrationResource_KeyValue` (`pkg/framework/objects/openai_integration`):
   ```
   Error: Unable to create OpenAI integration
     ...
     internal-server-error: <!doctype html>
     <html lang="en">
       ...
       <title>Uh oh! | dbt</title>
   ```

The next merges to `main` after each of these were both green with no code changes to either failing test, confirming these are environmental flakes — not test bugs.

## Root cause: control flow

In `pkg/dbt_cloud/client.go` (current `main`):

```go
for i := 0; i < c.MaxRetries; i++ {
    res, err := c.HTTPClient.Do(req)
    if err != nil { return nil, err }                      // (a) transport error → no retry
    defer res.Body.Close()                                 // (b) deferred inside loop
    body, err := io.ReadAll(res.Body)

    if res.StatusCode == 404 { … return … }                // (c) early returns
    if res.StatusCode == 400 { return … }
    if res.StatusCode == 401 { return … }
    if res.StatusCode == 403 { return … }
    if res.StatusCode == 500 {
        return nil, fmt.Errorf("internal-server-error: %s", body)
    }
    if res.StatusCode < 200 || res.StatusCode >= 300 {
        return nil, fmt.Errorf("unexpected status code %d: %s, URL: %s", res.StatusCode, body, req.URL)
    }

    if err == nil {
        return body, nil
    } else {
        if isErrorRetriable(res.StatusCode, c.RetriableStatusCodes) {  // (d) unreachable
            // exponential backoff + continue
        }
    }
    return nil, err
}
```

Mapping the May 26 failure (502) onto this:
- `res.StatusCode == 500` is false → fall through
- `res.StatusCode < 200 || res.StatusCode >= 300` is true → returns `unexpected status code 502: …` with no retry.

Mapping the April 26 failure (500):
- `res.StatusCode == 500` is true → returns `internal-server-error: …` with no retry.

In both cases the existing `isErrorRetriable` block (d) is never executed, because (c)'s early returns short-circuit out of the loop body. The block is only reachable when `err != nil` after `io.ReadAll` AND the status was 2xx — essentially never.

The provider already plumbs sensible defaults for `retriable_status_codes` (in `pkg/provider/framework_provider.go` they default to `["429", "500", "502", "503", "504"]`) and the README/schema documents that behavior, so configuration intent is correct — the implementation just doesn't honor it.

Secondary issues in the same function:

- `defer res.Body.Close()` inside the loop accumulates deferred closes across iterations instead of closing each response. Cosmetic given `io.ReadAll` drains the body, but still wrong.
- `res, err := c.HTTPClient.Do(req)` shadows the outer `err`, so the function's final `return nil, fmt.Errorf("max retries reached for request %s: %w", req.URL, err)` always wraps a `nil` (or pre-loop) `err`.
- Transport-level failures (DNS, connection reset, dial timeout) are also not retried.

## Fix (proposed)

1. Hoist the retry decision above the per-status terminal returns so any retriable status (defaults: `408, 429, 500, 502, 503, 504`, plus anything in `retriable_status_codes`) goes through the existing exponential-backoff loop.
2. Treat transport-level errors (`HTTPClient.Do` returning `err`) as retriable too.
3. After retries are exhausted, surface the same terminal error string the call site would have seen before this fix (`internal-server-error: …`, `unexpected status code …`, etc.) so existing callers that match on these substrings (e.g. resource Read/Create paths) keep working.
4. Fix the deferred-close-in-loop and shadowed `err` while we're in there.

A unit-tested implementation is up in PR (will link in this issue once opened).

## Impact

- Eliminates the repeated post-merge / scheduled `test-acceptance` flake class without paying the cost of a workflow-level rerun (re-running the entire ~5-min suite to fix one bad request).
- Makes the documented `retriable_status_codes` / `max_retries` provider settings actually work — today they are effectively dead config.
- Reduces transient-failure noise in real customer `terraform plan` / `apply` runs against `cloud.getdbt.com`.

## Out of scope

- Workflow-level retry on `make test-acceptance` (could be added as belt-and-suspenders, but the right fix lives in the client).
- Tightening the post-merge concurrency / matrix beyond what's already there (`concurrency.group` already serializes acceptance runs against the shared dbt Cloud account).

will-sargent-dbtlabs:
Resolves #695.

## Summary

`pkg/dbt_cloud/client.go::doRequestWithRetry` advertises retry-with-exponential-backoff via `MaxRetries`, `RetryIntervalSeconds`, and `RetriableStatusCodes` (provider config: `max_retries`, `retry_interval_seconds`, `retriable_status_codes`, defaulting to `[\"429\",\"500\",\"502\",\"503\",\"504\"]`). In practice the retry branch was unreachable: per-status `if res.StatusCode == … { return … }` blocks short-circuited every non-2xx response before the retry decision ran. Transient 5xx/429 responses from `cloud.getdbt.com` therefore surfaced as one-shot Terraform diagnostics on the very first attempt.

This change reorganizes the loop so any retriable status (defaults: `408, 429, 500, 502, 503, 504`, plus user-configured `retriable_status_codes`) and any transport-level failure now routes through the existing exponential-backoff path. After retries are exhausted, the surfaced error string is unchanged for back-compat with callers matching on \"internal-server-error:\", \"resource-not-found\", \"unauthorized:\", \"forbidden:\", and \"unexpected status code\" prefixes.

## Tracing the bug

Two recent post-merge / scheduled `unit (test-acceptance)` failures on `main`, no shared test, no shared package — only a shared symptom: a transient 5xx surfaced verbatim from the dbt Cloud API.

**2026-05-26**, run [26459657437](https://github.com/dbt-labs/terraform-provider-dbtcloud/actions/runs/26459657437), `TestAccDbtCloudPartialNotificationResource`:
\`\`\`
Error: Error Reading dbt Cloud Project
  ...
  Could not read project ID 543791: unexpected status code 502: <html>
  <head><title>502 Bad Gateway</title></head>
  ...
  , URL:
  https://cloud.getdbt.com/api/v3/accounts/***/projects/543791/?include_related=[freshness_job_id,docs_job_id]
\`\`\`

**2026-04-26** (scheduled), run [24947744549](https://github.com/dbt-labs/terraform-provider-dbtcloud/actions/runs/24947744549), `TestAccDbtCloudOpenAIIntegrationResource_OpenAI` and `_KeyValue`:
\`\`\`
Error: Unable to create OpenAI integration
  ...
  internal-server-error: <!doctype html>
  <html lang="en">
    ...
    <title>Uh oh! | dbt</title>
\`\`\`

The next merges to `main` after each were green with no test changes — confirming environmental flakes, not test bugs.

### Root cause

The previous `doRequestWithRetry` loop body looked like this (simplified):

\`\`\`go
for i := 0; i < c.MaxRetries; i++ {
    res, err := c.HTTPClient.Do(req)
    if err != nil { return nil, err }                      // (a) transport error → no retry
    defer res.Body.Close()                                 // (b) deferred inside loop
    body, err := io.ReadAll(res.Body)

    if res.StatusCode == 404 { … return … }                // (c) per-status early returns
    if res.StatusCode == 400 { return … }
    if res.StatusCode == 401 { return … }
    if res.StatusCode == 403 { return … }
    if res.StatusCode == 500 {
        return nil, fmt.Errorf(\"internal-server-error: %s\", body)
    }
    if res.StatusCode < 200 || res.StatusCode >= 300 {
        return nil, fmt.Errorf(\"unexpected status code %d: %s, URL: %s\", res.StatusCode, body, req.URL)
    }

    if err == nil {
        return body, nil
    } else {
        if isErrorRetriable(res.StatusCode, c.RetriableStatusCodes) {  // (d) unreachable
            // backoff + continue
        }
    }
    return nil, err
}
\`\`\`

Mapping the May 26 failure (502):
- `res.StatusCode == 500` is false → fall through.
- `res.StatusCode < 200 || res.StatusCode >= 300` is true → returns `unexpected status code 502: …` with no retry.

Mapping the April 26 failure (500):
- `res.StatusCode == 500` is true → returns `internal-server-error: …` with no retry.

In both cases the retry block (d) is unreachable because the per-status early returns (c) bail out first. (d) is only reached when `err != nil` from `io.ReadAll` AND status is 2xx — essentially never.

Secondary issues fixed alongside:
- `defer res.Body.Close()` inside the loop accumulates deferred closes across iterations instead of closing per-iteration.
- `res, err := c.HTTPClient.Do(req)` shadows the outer `err`, so the function's exhausted-retries `return nil, fmt.Errorf(\"max retries reached for request %s: %w\", req.URL, err)` always wraps a `nil`-or-stale `err`.
- Transport-level failures (DNS, connection reset, dial timeout) were not retried.

## Fix

`doRequestWithRetry` now decides retriability *before* mapping the status into a terminal error. The mapping itself moves into a small helper, `attemptRequest`, which preserves every existing error-message format. This is the minimal change that makes the documented behavior (retriable_status_codes defaulting to `[429, 500, 502, 503, 504]`) actually work.

Key behavior changes:
- `408, 429, 500, 502, 503, 504` retry by default; **501 stays terminal** (intentionally — Not Implemented is a permanent missing feature, not a transient failure).
- User-configured `retriable_status_codes` continue to take precedence; the default list is additive, not overriding.
- Transport-level errors retry once per attempt as well.
- After retries exhaust, the error string is byte-identical to today's terminal error (locked in by tests).
- 4xx terminal codes (400/401/403/404) still surface immediately on the first attempt — no wasted retry budget on permission/validation errors.
- `isErrorRetriable` is preserved as a thin wrapper that defers to the new `isHTTPCodeRetriable`, so any external caller (none in this repo, but harmless) still works.

## Test plan

- [x] `go build ./pkg/dbt_cloud/...`
- [x] `go vet ./pkg/dbt_cloud/...`
- [x] `go test -count=1 ./pkg/dbt_cloud/...` (existing + new)
- [x] `go test -count=1 ./...` — full unit suite (no `TF_ACC`) green
- [x] `TestDoRequestWithRetry_RetriesTransient5xxAnd429` covers `408, 429, 500, 502, 503, 504` retry-then-success
- [x] `TestDoRequestWithRetry_PreservesTerminalErrorOn500Exhaustion` locks `internal-server-error:` substring
- [x] `TestDoRequestWithRetry_PreservesTerminalErrorOn502Exhaustion` locks `unexpected status code 502` substring
- [x] `TestDoRequestWithRetry_DoesNotRetry4xx` (401 / 403 / 400) — exactly one attempt
- [x] `TestDoRequestWithRetry_HonorsConfiguredRetriableStatusCodes` (418 via configured list)
- [x] `TestDoRequestWithRetry_RetriesTransportError` (hijacked-and-closed connection)
- [x] `TestDoRequestWithRetry_DisableRetryStillReturnsSuccess`
- [x] `TestIsHTTPCodeRetriable` table tests
- [x] `TestIsErrorRetriable_BackwardsCompatible` (legacy helper still works)
- [ ] CI `unit (test-acceptance)` — passes against the live dbt Cloud API.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changes/unreleased/Fixes-20260526-210625.yaml
+++ b/.changes/unreleased/Fixes-20260526-210625.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: API client now actually retries transient HTTP failures (500, 502, 503, 504, 408, 429 and any user-configured `retriable_status_codes`). Previously `doRequestWithRetry` short-circuited every non-2xx response with an early return before the retry/backoff branch could fire, so transient `cloud.getdbt.com` 5xx responses surfaced as one-shot Terraform diagnostics on the first attempt. Network/transport-level errors are now retried as well (resolves #695).
+time: 2026-05-26T21:06:25.000000+00:00
--- a/pkg/dbt_cloud/client.go
+++ b/pkg/dbt_cloud/client.go
@@ -178,114 +178,172 @@ func NewClient(account_id *int64, token *string, host_url *string, maxRetries *i
 	return &c, nil
 }
 
-func (c *Client) doRequestWithRetry(req *http.Request) ([]byte, error) {
-	var err error
+// defaultRetriableHTTPCodes lists the HTTP status codes the client retries by
+// default when the provider has not supplied an explicit retriable_status_codes
+// list. They cover the transient upstream / proxy / rate-limit failures that
+// have caused real customer apply errors and post-merge CI flakes:
+//
+//   - 408 Request Timeout
+//   - 429 Too Many Requests
+//   - 500 Internal Server Error
+//   - 502 Bad Gateway
+//   - 503 Service Unavailable
+//   - 504 Gateway Timeout
+//
+// 501 Not Implemented is intentionally excluded because it indicates a
+// permanent missing feature rather than a transient failure.
+var defaultRetriableHTTPCodes = []int{408, 429, 500, 502, 503, 504}
 
-	// This is needed in case the provider code wants to do retries but the provider config is set to disable retries
+func (c *Client) doRequestWithRetry(req *http.Request) ([]byte, error) {
+	// Provider config may want to disable retries even when client code asks
+	// for them. Floor MaxRetries at 1 so we always make at least one attempt.
 	if c.DisableRetry || c.MaxRetries <= 0 {
 		c.MaxRetries = 1
 	}
 
 	setRequestHeaders(req, c.Token)
 
-	for i := 0; i < c.MaxRetries; i++ {
-		res, err := c.HTTPClient.Do(req)
-		if err != nil {
-			return nil, err
-		}
-
-		defer res.Body.Close()
-
-		body, err := io.ReadAll(res.Body)
-
-		// Handle 404 errors - check if it's a true not found or a permissions issue
-		if res.StatusCode == 404 {
-			isResourceNotFound, apiErr, parseErr := parseAPIError(body)
-			if parseErr != nil {
-				// If we can't parse the error, return a generic 404
-				return nil, fmt.Errorf("resource-not-found (status 404): URL: %s, Response: %s", req.URL, body)
-			}
-
-			if isResourceNotFound {
-				// Check if the error message mentions permissions - this is a common pattern in dbt Cloud API
-				userMsg := strings.ToLower(apiErr.Status.UserMessage)
-				if strings.Contains(userMsg, "permission") || strings.Contains(userMsg, "proper permissions") {
-					return nil, fmt.Errorf("resource-not-found-permissions: The resource was not found, but this may be due to insufficient permissions. The API token may not have access to this resource or the environment it belongs to.\n\nStatus: 404\nURL: %s\nMessage: %s", req.URL, apiErr.Status.UserMessage)
-				}
-
-				// For GET requests or DELETE operations, this is typically a legitimate not-found
-				// (DELETE gets 404 when resource already deleted, which is fine)
-				if req.Method == "GET" || req.Method == "DELETE" {
-					return nil, fmt.Errorf("resource-not-found: %s", req.URL)
-				}
-
-				// For POST/PUT on non-permission 404s, provide additional context
-				// This helps with update/create operations that fail due to permissions
-				return nil, fmt.Errorf("resource-not-found: The resource was not found. If you are updating a resource, this may indicate insufficient permissions.\n\nStatus: 404\nURL: %s\nMessage: %s", req.URL, apiErr.Status.UserMessage)
-			}
-		}
-
-		if res.StatusCode == 400 {
-			return nil, fmt.Errorf("resource-not-found: %s", body)
-		}
-
-		// Handle permission errors (401 Unauthorized, 403 Forbidden)
-		if res.StatusCode == 401 {
-			return nil, fmt.Errorf("unauthorized: The API token does not have permission to access this resource. Status: 401, URL: %s, Response: %s", req.URL, body)
-		}
-
-		if res.StatusCode == 403 {
-			return nil, fmt.Errorf("forbidden: The API token does not have permission to perform this action. This may be due to environment-level permissions or other access restrictions. Status: 403, URL: %s, Response: %s", req.URL, body)
-		}
-
-		if res.StatusCode == 500 {
-			return nil, fmt.Errorf("internal-server-error: %s", body)
-		}
-
-		// Check for other non-2xx status codes
-		if res.StatusCode < 200 || res.StatusCode >= 300 {
-			return nil, fmt.Errorf("unexpected status code %d: %s, URL: %s", res.StatusCode, body, req.URL)
-		}
-
-		if err == nil {
+	var lastErr error
+	for attempt := 0; attempt < c.MaxRetries; attempt++ {
+		body, statusCode, attemptErr := c.attemptRequest(req)
+		if attemptErr == nil {
 			return body, nil
-		} else {
-			if isErrorRetriable(res.StatusCode, c.RetriableStatusCodes) {
-				waitDuration := time.Duration(c.RetryIntervalSeconds) * time.Second
-				// Exponential backoff
-				if i > 0 {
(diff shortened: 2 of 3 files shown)


will-sargent-dbtlabs:
@adrianburusdbt @tudor-scurtu This was to fix some of the intermittent merge deploy runs that were failing

adrianburusdbt (approved):

PR #697, opened 2026-05-28
Fix update for BQ global connection

### LINKED ISSUE #685: BigQuery v1 global connections cannot be updated — adapter_version in PATCH payload
## Describe the bug

`dbtcloud_global_connection` updates fail for BigQuery v1 connections because the provider sends `adapter_version` in PATCH requests, which the dbt Cloud API rejects as an invalid field.

The `UpdateWithLatestAdapter()` function in `pkg/dbt_cloud/global_connection.go` includes `adapter_version` in the payload:

```go
// lines 204-218
func (c *GlobalConnectionClient[T]) UpdateWithLatestAdapter(...) {
    payload := globalConnectionPayload[T]{
        GlobalConnectionCommon: common,
        AccountID:              int64(c.AccountID),
        AdapterVersion:         &av,  // <-- Should not be sent on PATCH
        Config:                 config,
    }
```

The `Update()` function (lines 220-230) correctly omits this field and works fine. The issue is that BigQuery v1 connections route through `UpdateWithLatestAdapter()`.

This makes BigQuery v1 global connections effectively **immutable** after creation — any in-place update fails.

## Error message

```
Error: Unable to update global connection

"developer_message":"Invalid request data: adapter_version: extra fields not permitted"
```

## Resource configuration

```hcl
resource "dbtcloud_global_connection" "bigquery_wif" {
  name                   = "my-bigquery-connection"
  oauth_configuration_id = 341

  bigquery = {
    gcp_project_id              = "my-gcp-project"
    timeout_seconds             = 300
    private_key_id              = "placeholder"
    private_key                 = "placeholder"
    client_email                = "sa@my-gcp-project.iam.gserviceaccount.com"
    client_id                   = "123456789"
    auth_uri                    = "https://accounts.google.com/o/oauth2/auth"
    token_uri                   = "https://oauth2.googleapis.com/token"
    auth_provider_x509_cert_url = "https://www.googleapis.com/oauth2/v1/certs"
    client_x509_cert_url        = "https://www.googleapis.com/robot/v1/metadata/x509/..."
  }
}
```

Changing any attribute (e.g., `name`, `oauth_configuration_id`, `timeout_seconds`) and running `terraform apply` triggers the error.

## Expected behavior

The connection should update in-place, the same as other connection types that use the `Update()` code path.

## Config

- dbt Cloud provider version: **v1.10.3** (also affects v1.9.1, v1.10.0)

## Additional context

**Workaround:** The only current workaround is to destroy and recreate connections for any change. Adding `lifecycle { ignore_changes = [...] }` doesn't help because the update code path still sends the rejected field.

**Suggested fix:** Remove `AdapterVersion: &av` from `UpdateWithLatestAdapter()` to match the working `Update()` function:

```go
payload := globalConnectionPayload[T]{
    GlobalConnectionCommon: common,
    AccountID:              int64(c.AccountID),
    // AdapterVersion omitted - API rejects it on PATCH
    Config:                 config,
}
```

### LINKED ISSUE #686: Add WIF authentication fields to dbtcloud_bigquery_credential
When managing BigQuery connections with Workload Identity Federation (WIF), the `dbtcloud_global_connection` resource supports WIF configuration, but `dbtcloud_bigquery_credential` does not expose the credential-level fields required to complete WIF setup.
This means credentials created via Terraform default to service account JSON auth, even when the connection is configured for WIF. The credential must be manually patched via direct API calls to enable WIF authentication.
Add WIF-related fields to `dbtcloud_bigquery_credential`:
- **`auth_type`** — authentication method, one of `service-account-json`, `oauth-secrets`, or `external-oauth-wif`
- **`workload_pool_provider_path`** — the fully qualified WIF provider resource name (e.g., `//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/my-pool/providers/my-provider`)
These map to the `credential_details.fields` structure in the v3 API:
```json
{
  "credential_details": {
    "fields": {
      "auth_type": { "value": "external-oauth-wif" },
      "workload_pool_provider_path": { "value": "//iam.googleapis.com/..." }
    }
  }
}
```
## Describe alternatives you've considered
**Direct API calls via shell script** — We currently run an idempotent shell script after `terraform apply` that POSTs to `/api/v3/accounts/{account_id}/projects/{project_id}/credentials/{credential_id}/` with the full `credential_details.fields` payload. This works but breaks the Terraform-as-source-of-truth model and adds operational complexity.
**`dbtcloud_extended_attributes`** — We tried attaching `auth_type` and `workload_pool_provider_path` via extended attributes on the profile, but these map to `profiles.yml` runtime keys and do not set the underlying credential's `auth_type` in the API.
## Additional context
Related: #584 added WIF support to `dbtcloud_global_connection`. This request completes WIF support on the credential side.

adrianburusdbt:
## Summary

Fixes #685: BigQuery v1 global connections (`use_latest_adapter = true`) were effectively immutable. The provider sent `adapter_version` in every PATCH, and the dbt Cloud API rejects that field (`adapter_version: extra fields not permitted`), so renaming a v1 connection — or changing any other mutable attribute — failed.

The dbt Cloud API contract is that adapter version is set at create and immutable thereafter (`AccountConnectionUpdateBody` schema in `sinter/api/v3/account_connections/account_connections_schemas.py` has no `adapter_version` field, and there's a backend regression test asserting the 400). So the right fix is provider-side: stop sending the field on PATCH, and warn users earlier if they try to change it.

## Changes

- **`pkg/dbt_cloud/global_connection.go`** — removed `UpdateWithLatestAdapter`. Once `adapter_version` is dropped from the payload it's identical to `Update`.
- **`pkg/framework/objects/global_connection/resource.go`** — v0 and v1 BigQuery updates now share the same PATCH call. Added a `ModifyPlan` check that rejects `use_latest_adapter` changes at plan time with an actionable message pointing at `terraform apply -replace=<address>`. The existing runtime check at line 897 stays as defense in depth.
- **`pkg/framework/objects/global_connection/resource_acceptance_test.go`** — new `TestAccDbtCloudGlobalConnectionBigQueryV1UpdateIssue685` covering: create v1, rename (the reporter's scenario), change another v1 attribute, and the plan-time guard against flipping `use_latest_adapter`.
- **`test-scenarios/issue-685-bigquery-v1-update/`** — manual repro workspace with a README.
- Two changelog entries under `.changes/unreleased/`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `TF_ACC=1 go test -run TestAccDbtCloudGlobalConnectionBigQueryV1UpdateIssue685 -timeout 30m ./pkg/framework/objects/global_connection/...`
- [x] Manual repro per `test-scenarios/issue-685-bigquery-v1-update/README.md`
- [ ] (Optional) Revert `pkg/dbt_cloud/global_connection.go` and re-run the acceptance test — step 2 should fail with `adapter_version: extra fields not permitted`, confirming the test catches the regression

### DIFF
--- a/.changes/unreleased/Fixes-20260528-093810.yaml
+++ b/.changes/unreleased/Fixes-20260528-093810.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: Fix update for BQ global connection
+time: 2026-05-28T09:38:10.215562+03:00
--- a/pkg/dbt_cloud/global_connection.go
+++ b/pkg/dbt_cloud/global_connection.go
@@ -202,26 +202,6 @@ func (c *GlobalConnectionClient[T]) createGlobalConnection(
 	return &resp.Data, nil
 }
 
-func (c *GlobalConnectionClient[T]) UpdateWithLatestAdapter(
-	connectionID int64,
-	common GlobalConnectionCommon,
-	config T,
-	av string,
-) (*GlobalConnectionCommon, *T, error) {
-
-	buffer := new(bytes.Buffer)
-	enc := json.NewEncoder(buffer)
-
-	payload := globalConnectionPayload[T]{
-		GlobalConnectionCommon: common,
-		AccountID:              int64(c.AccountID),
-		AdapterVersion:         &av,
-		Config:                 config,
-	}
-
-	return updateGlobalConnection(enc, payload, c, connectionID, buffer)
-}
-
 func (c *GlobalConnectionClient[T]) Update(
 	connectionID int64,
 	common GlobalConnectionCommon,
--- a/pkg/framework/objects/global_connection/resource.go
+++ b/pkg/framework/objects/global_connection/resource.go
@@ -94,6 +94,25 @@ func (r globalConnectionResource) ModifyPlan(
 		}
 	}
 
+	// dbt Cloud does not allow changing the adapter version of an existing connection
+	// (the PATCH endpoint rejects `adapter_version`). Surface this at plan time with a
+	// clear, actionable message so the user sees it before `terraform apply` starts
+	// mutating other attributes. The Update path retains a defensive runtime check.
+	if plan.BigQueryConfig != nil && state.BigQueryConfig != nil &&
+		plan.BigQueryConfig.UseLatestAdapter.ValueBool() != state.BigQueryConfig.UseLatestAdapter.ValueBool() {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("bigquery").AtName("use_latest_adapter"),
+			"Adapter version cannot be changed on an existing connection",
+			"dbt Cloud does not support changing the BigQuery adapter version "+
+				"(`use_latest_adapter`) on an existing global connection. To switch between "+
+				"the legacy (bigquery_v0) and latest (bigquery_v1) adapter, recreate the "+
+				"connection — for example, run `terraform apply -replace=<resource address>` "+
+				"(see https://developer.hashicorp.com/terraform/cli/commands/plan#replace-address). "+
+				"Note that recreating the connection will issue a new connection ID, so any "+
+				"resources referencing it will also be updated.",
+		)
+	}
+
 }
 
 func (r *globalConnectionResource) Read(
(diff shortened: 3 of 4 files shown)


victorasu (approved):
Banger :shipit:

PR #701, opened 2026-06-03
State changes

### LINKED ISSUE #702: Support for dbt state

adrianburusdbt:
## What

Adds **dbt State** support to `dbtcloud_job` by accepting `dbt_state` in `cost_optimization_features` — the migration path from Select All Optimizations (SAO).

## Why the change is more than a new enum value

`cost_optimization_features` was never sent to the API; it was bridged through the `force_node_selection` boolean. A boolean can only express SAO on/off, so it can't carry a distinct feature like dbt State.

This PR makes `cost_optimization_features` a **first-class field** — sent to and read back from the dbt platform API. State is reconciled from the API response, which derives `force_node_selection` from the set server-side.

## Behavior

| Config | Result |
| --- | --- |
| `["state_aware_orchestration"]` | SAO on → `force_node_selection = false` |
| `["dbt_state"]` | dbt State on → `force_node_selection = false` |
| `[]` / unset | `force_node_selection = true` |

> [!NOTE]
> A new validator rejects mixing `dbt_state` with other features, because the API collapses such a set to `["dbt_state"]`.

> [!IMPORTANT]
> - `dbt_state` is environment-independent — no Fusion version or prod/staging env required — but needs dbt State enabled on the account.
> - Not supported on CI or Merge jobs.

## Testing

- Validator unit test (runs in CI).
- Acceptance tests (`TF_ACC=1`) for SAO and dbt State, each asserting the API-derived `force_node_selection`.
- Manual config under `examples/local-test-job-dbt-state/`.


pkg/framework/objects/job/resource_acceptance_test.go
```diff
@@ -948,10 +948,17 @@ func TestAccDbtCloudJobResourceIntervalCron(t *testing.T) {
 // back to true, which makes Terraform see an inconsistent result after apply
 // (plan: ["state_aware_orchestration"], actual: []).
 //
-// This test is skipped by default because the default acceptance test environment
-// uses dbt_version="latest" (non-Fusion) and we cannot assume SAO enforcement.
+// It is gated behind DBT_CLOUD_ACC_TEST_SAO_ENABLED so it only runs against an
```
vladmld: Where is this `DBT_CLOUD_ACC_TEST_SAO_ENABLED` happening? 🤔 
Isn't it gated behind: `acctest_config.IsCI()`?

Same for `TestAccDbtCloudJobDbtStateCostOptimizationFeature`.

vladmld (approved):
Approved with a question. 👍🏼

PR #707, opened 2026-06-17
Handle other / scheduled job types on api response

adrianburusdbt:
## Summary

Fixes a `Provider produced inconsistent result after apply` error on `dbtcloud_job` when `job_type` is set to `scheduled` but the dbt platform API derives `other` (or vice versa).

`scheduled` and `other` are not independently settable values — the API derives them from whether a job has an active schedule trigger. When a config sets `job_type = "scheduled"` on a job whose triggers don't include an active schedule (common for deactivated jobs, or jobs triggered only via job completion / API), the API stores it as `other`. Because the provider overwrote the planned value with the API response after apply, Terraform saw the known config value (`scheduled`) change to `other` and raised:

```

When applying changes to dbtcloud_job.x, provider produced an unexpected new
value: .job_type: was cty.StringVal("scheduled"), but now cty.StringVal("other").
```

## What changed

- **Treat `scheduled` ↔ `other` as the interchangeable derived pair they are.** A new `reconcileJobType` helper preserves the configured/prior label when both the desired and API values fall in that soft pair, while keeping `ci` / `merge` / `adaptive` authoritative from the API. Applied at all three sites that set `job_type` from an API response: create, update, and read. Read is included so refreshed state doesn't drift back to `other` and produce a perpetual diff.
- **Corrected the `job_type` attribute description.** It previously claimed the value could "enforce" the job type including `scheduled`, which is inaccurate — `scheduled`/`other` are inferred and cannot be enforced. Docs regenerated accordingly.
- **Added a unit test** (`reconcileJobType`) covering soft-pair preservation, API-authoritative `ci`/`merge`, and null/empty handling.

## Behavior

| `job_type` in config | API derives | Before | After |
|---|---|---|---|
| `scheduled` | `other` | error: inconsistent result | keeps `scheduled`, applies cleanly |
| `other` | `scheduled` | drift on next plan | keeps `other` |
| `ci` / `merge` / `adaptive` | same | unchanged | unchanged (API authoritative) |

The rigid transition rules (`ci`/`adaptive` force replacement) are untouched.

## Testing

- `go test ./pkg/framework/objects/job/` — new `TestReconcileJobType` table test passes alongside existing `TestValidateJobTypeChange`.

> Note: build/tests should be run by a reviewer or CI; they were not executed in the authoring environment.

## Context

Surfaced during a customer account migration using `dbtcloud-terraforming`, which exports an explicit `job_type = "scheduled"` for jobs regardless of their triggers. A follow-up on the terraforming side (not exporting non-enforceable `job_type`, and fixing Azure DevOps repository exports) is tracked separately.


victorasu (approved):
LGTM! I'm curious about whether the plan/state drift is an issue here, I know handling the change of plan/state diffs is tricky.

:shipit:

PR #710, opened 2026-06-22
Fix schema validation for sl creds

adrianburusdbt:
### Bug fix — relative path expressions
All `ConflictsWith` validators in the Snowflake, Databricks, Postgres, and Redshift credential schemas now use `path.MatchRelative().AtParent().AtName(...)` instead of `path.MatchRoot(...)`. Relative paths work correctly whether the schema is used at the resource root or nested inside another attribute.

### New helper validator
`pkg/helper/prefer_write_only_validator.go` introduces `PreferWriteOnlyAttributeValidator` to replace `stringvalidator.PreferWriteOnlyAttribute(...)` from the framework-validators library. The built-in validator takes an absolute path expression (same root issue); this replacement resolves the write-only sibling via `req.Path.ParentPath().AtName(...)`.

### Regression test
`TestSnowflakeSLCredential_NestedSchemaValidatesPlan` is an offline unit test that runs `PlanOnly` steps for both password and keypair auth configs against a mock server — no live credentials required. It reproduces the original failure and guards against regressions.

## Testing

- [x] Existing credential acceptance tests pass
- [x] New unit test `TestSnowflakeSLCredential_NestedSchemaValidatesPlan` passes
- [x] `dbtcloud_snowflake_semantic_layer_credential` plan no longer errors with `Invalid Path Expression for Schema`

### DIFF
--- a/.changes/unreleased/Fixes-20260622-165954.yaml
+++ b/.changes/unreleased/Fixes-20260622-165954.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: Fix handling of wo attributes and sl credentials
+time: 2026-06-22T16:59:54.291212+03:00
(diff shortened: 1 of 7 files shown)


tomighita (approved):
Lgtm! 🕶️

PR #720, opened 2026-08-03
fix: spurious in-place update on dbtcloud_postgres_credential.id

### LINKED ISSUE #719: dbtcloud_postgres_credential: id always shows (known after apply) when using password_wo
## Description

Any `dbtcloud_postgres_credential` resource that uses `password_wo` (the write-only password argument) shows an in-place update on **every** `terraform plan`, even when nothing has actually changed. The only visible attribute in the diff is `id`:

```
  # dbtcloud_postgres_credential.example will be updated in-place
!~  resource "dbtcloud_postgres_credential" "example" {
!~      id = "*************" -> (known after apply)
      # (11 unchanged attributes hidden)
    }
```

This happens on a clean, unmodified config with no other drift — confirmed by comparing against the live dbt Cloud API, where all 11 hidden attributes genuinely match.

## Root cause (from reading the provider source)

1. `password_wo` has no companion `ModifyPlan` logic in [`resource.go`](https://github.com/dbt-labs/terraform-provider-dbtcloud/blob/765cc5634133d1d6c58a9b7cebe673493392b23f/pkg/framework/objects/postgres_credential/resource.go). The docstring for `password_wo_version` says it's required "to trigger updates," but nothing in the resource actually reads/compares it to decide whether an update is needed — so the resource is unconditionally considered for update whenever `password_wo` is set, regardless of `password_wo_version`.
2. In [`schema.go`](https://github.com/dbt-labs/terraform-provider-dbtcloud/blob/765cc5634133d1d6c58a9b7cebe673493392b23f/pkg/framework/objects/postgres_credential/schema.go#L57-L60), the `id` attribute is `Computed` with no `PlanModifiers`:

   ```go
   "id": resource_schema.StringAttribute{
       Computed:    true,
       Description: "The ID of this resource. Contains the project ID and the credential ID.",
   },
   ```

   Compare to `credential_id` a few lines below, which correctly has `UseStateForUnknown()`:

   ```go
   "credential_id": resource_schema.Int64Attribute{
       Computed:    true,
       ...
       PlanModifiers: []planmodifier.Int64{
           int64planmodifier.UseStateForUnknown(),
       },
   },
   ```

   Since `id` is just `fmt.Sprintf("%d:%d", project_id, credential_id)` — both of which are already stable and known — it should never need to be recomputed as unknown. The missing plan modifier means it gets marked unknown any time the resource is touched for update, which surfaces as the only visible line in the diff.

## Impact

- Every `terraform apply` re-sends the plaintext password to the dbt Cloud API even when it hasn't changed.
- Noisy plans make it harder to spot genuine drift for these resources, since they always show as needing an update.
- This likely affects the other `_wo\" resources in the provider using the same write-only pattern (e.g. `dbtcloud_snowflake_credential`'s `private_key_wo`) if they share the same gap — I haven't audited those, but the underlying pattern is presumably shared.

## Expected behavior

- `terraform plan` should show no changes for a `dbtcloud_postgres_credential` resource using `password_wo` when `password_wo_version` is unchanged and no other attribute differs.
- At minimum, `id` should carry `UseStateForUnknown()` so it doesn't spuriously appear in the diff.

## Terraform / provider versions

- `dbtcloud` provider: `1.11.1`
- Terraform: `>= 1.11` (required for write-only argument support)

## Reproduction

```hcl
resource "dbtcloud_postgres_credential" "example" {
  default_schema      = "target"
  is_active           = true
  num_threads          = 4
  password_wo          = var.pg_password
  password_wo_version  = 1
  project_id           = dbtcloud_project.example.id
  target_name          = "default"
  username             = var.pg_username
}
```

Run `terraform plan` twice in a row with no config changes in between — the second plan still shows `id` as `(known after apply)` and the resource as "will be updated in-place."

dbtagarovat:
## Summary
- `id` in the `dbtcloud_postgres_credential` resource schema was missing `UseStateForUnknown()`, unlike `credential_id` in the same schema and the `id` attribute of every other resource in this provider.
- Since `id` is deterministically derived from `project_id` and `credential_id` (both already stable/known), it never needs to be recomputed as unknown. The missing plan modifier caused `terraform plan` to always show a no-op in-place update whenever the resource used `password_wo` — the only visible line in the diff being `id = "..." -> (known after apply)`.
- Fix: add `UseStateForUnknown()` to the `id` attribute, matching the rest of the provider.

## Verification
Also confirmed live against a real dbt Cloud backend (personal devspace):
- Built the pre-fix (`main`) provider, applied a `postgres_credential` using `password_wo`, then ran `plan` again with no config changes — reproduced the exact bug (`id = "160:227" -> (known after apply)`, planned as an update).
- Rebuilt with this fix, ran `plan` again on the same state — confirmed via the plan JSON diff that `id` is now identical before/after with no change proposed for it.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/framework/objects/postgres_credential/...` (acceptance tests skip without `TF_ACC`, package compiles and unit-level checks pass)

Closes #719

### DIFF
--- a/.changes/unreleased/Fixes-20260803-132946.yaml
+++ b/.changes/unreleased/Fixes-20260803-132946.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: Add `UseStateForUnknown()` to `dbtcloud_postgres_credential`'s `id` attribute so `terraform plan` no longer shows a spurious in-place update on every run when using `password_wo` (#719).
+time: 2026-08-03T13:29:46.014118+03:00
--- a/pkg/framework/objects/postgres_credential/behavior_test.go
+++ b/pkg/framework/objects/postgres_credential/behavior_test.go
@@ -1,0 +1,72 @@
+package postgres_credential
+
+import (
+	"context"
+	"testing"
+
+	resource_schema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// TestIDAttribute_UseStateForUnknown is a regression test for #719: the "id"
+// attribute had no PlanModifiers, so it was always recomputed as unknown
+// whenever the resource was touched for an update (e.g. one triggered by
+// bumping password_wo_version), even though id is deterministically derived
+// from project_id and credential_id and never actually changes. That surfaced
+// as a spurious "id = ... -> (known after apply)" in-place update on every
+// terraform plan.
+func TestIDAttribute_UseStateForUnknown(t *testing.T) {
+	attr, ok := PostgresResourceSchema.Attributes["id"].(resource_schema.StringAttribute)
+	if !ok {
+		t.Fatalf("id attribute is not a resource_schema.StringAttribute")
+	}
+	if len(attr.PlanModifiers) == 0 {
+		t.Fatalf("regression: id attribute has no plan modifiers, so terraform will always recompute it as unknown")
+	}
+
+	// Build a real prior state matching the full resource schema: the plan
+	// modifier bails out early if req.State.Raw is null, so a bare
+	// StringRequest{StateValue: ...} isn't enough to exercise it faithfully.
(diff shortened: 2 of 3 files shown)


tudor-scurtu (approved):
Thank you! :shipit:

PR #721, opened 2026-08-03
fix: bigquery_v1 global connection import and update-path adapter issues

### LINKED ISSUE #709: Import gap: `use_latest_adapter` is not set for imported bigquery_v1 connections, causing adapter-change errors and compounding timeout_seconds bug
Addendum: import/read does not set `use_latest_adapter` for bigquery_v1 connections — causes false 'Adapter version cannot be changed' errors and compounds timeout_seconds issue.

When importing a BigQuery v1 global connection (either via CLI `terraform import` or config-driven `import {}`), the provider's read/import path populates adapter_version = "bigquery_v1" but does not set `bigquery.use_latest_adapter` in state. This means state has `use_latest_adapter = null` while the connection is v1. If the config has `use_latest_adapter = true`, the provider's ModifyPlan guard treats this as a user-requested adapter change and errors with "Adapter version cannot be changed". If the config omits the flag, the plan treats the resource as v0, injects the v0-only default `timeout_seconds = 300` into the plan, and the Update path sends it to the API, which rejects it for BigqueryConnectionV1.

Proposed remediation steps to include in the same issue:

1. In readGeneric (pkg/framework/objects/global_connection/common.go), when reading a BigQuery connection, derive and set `state.BigQueryConfig.UseLatestAdapter` based on the adapter_version returned by the API (e.g., true if adapter_version starts with "bigquery_v1" or equals the latest adapter). This allows imported v1 connections to round-trip and avoids the ModifyPlan false positive.

2. In the Update path (pkg/framework/objects/global_connection/resource.go), guard sending `TimeoutSeconds` to the API: only include `warehouseConfigChanges.TimeoutSeconds` when the plan indicates the legacy adapter (use_latest_adapter == false). The Create path already does adapter-aware handling; mirror that behavior in Update.

3. Add regression tests for:
   - Importing a bigquery_v1 connection and then running a no-op plan/apply (should be no adapter-change errors).
   - Updating a bigquery_v1 connection with no timeout_seconds set should NOT PATCH `timeout_seconds`.

Rationale: these two changes together ensure imported v1 connections behave correctly in Terraform, and prevent the provider from sending v0-only fields to v1 API objects. They resolve both the immediate failure mode and prevent future surprises for other users using import blocks.

dbtagarovat:
## Summary
- `readGeneric` (shared by `Read` and `ImportState`) never set `BigQueryConfig.UseLatestAdapter`, so importing (or refreshing) a `bigquery_v1` `dbtcloud_global_connection` left `use_latest_adapter = null` in state. The `ModifyPlan` guard then compared that null (`false`) against the configured `true` and raised a false "Adapter version cannot be changed on an existing connection" error on a plain no-op plan.
- Fix: derive `UseLatestAdapter` from the adapter version returned by the API in `readGeneric`, mirroring what `Create` already does explicitly after creating a v1 connection.
- Also found and fixed a related issue in the same code path: `Update` unconditionally sent `timeout_seconds` whenever it differed from state, but `timeout_seconds` is a `bigquery_v0`-only field that the API rejects for `bigquery_v1`. `Create` already handles this correctly per-adapter; `Update` now does too.

## Verification
Also confirmed live against a real dbt Cloud backend (personal devspace):
- Built the pre-fix (`main`) provider, created a `bigquery_v1` global connection (`use_latest_adapter = true`), removed it from state, then ran `terraform import` on the same connection — reproduced the exact reported error: "Adapter version cannot be changed on an existing connection" on a plan with zero config changes.
- Rebuilt with this fix, repeated the same remove-and-reimport, then `plan` — the false error is gone; `use_latest_adapter` is no longer part of the diff.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/framework/objects/global_connection/...` (acceptance tests skip without `TF_ACC`, package compiles and unit-level checks pass)

Closes #709

### DIFF
--- a/.changes/unreleased/Fixes-20260803-132947.yaml
+++ b/.changes/unreleased/Fixes-20260803-132947.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: 'Fix import/refresh of `bigquery_v1` `dbtcloud_global_connection` resources: `use_latest_adapter` is now derived from the API''s adapter version instead of being left unset, avoiding a false "Adapter version cannot be changed" error, and `timeout_seconds` (a `bigquery_v0`-only field) is no longer sent on update for `bigquery_v1` connections (#709).'
+time: 2026-08-03T13:29:47.27689+03:00
--- a/pkg/framework/objects/global_connection/behavior_test.go
+++ b/pkg/framework/objects/global_connection/behavior_test.go
@@ -1,0 +1,366 @@
+package global_connection
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/dbt-labs/terraform-provider-dbtcloud/pkg/dbt_cloud"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// These are plain Go unit tests (no TF_ACC / live API needed) that exercise the
+// exact code paths fixed for GitHub issue #709: importing/refreshing a
+// bigquery_v1 dbtcloud_global_connection left use_latest_adapter unset in
+// state, which the ModifyPlan guard then misread as a user-requested adapter
+// change, and Update unconditionally sent the bigquery_v0-only
+// timeout_seconds field even for bigquery_v1 connections.
+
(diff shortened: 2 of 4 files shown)


tudor-scurtu (approved):
Thank you! :shipit:

claudiazi:
helloo!! are we merging this fix soon? :)

siyuan-liu-shawn:
Hi! We are meeting the same problem. When this PR can be merged?

claudiazi:
Helloo!!! Any plans for this fix?

PR #722, opened 2026-08-03
fix: make dbtcloud_notification_setting delete idempotent on 404

### LINKED ISSUE #718: `notification_settings` delete sometimes returns 404 for no reason
## Describe the bug
When deleting notification settings, the server sometimes returns a 404 error code during the operation even though it deletes the resource from the DB.

## Error message
Unclear.

## Resource configuration
Unclear.

## Expected behavior
Deleting a notification setting would always return 2xx codes when succeeding, 4xx when the setting is already removed and 5xx when the server encounters an unexpected error.

Unclear.

## Additional context
An fix for this by making the DELETE operation idempotent was attempted in [dbt-cloud](https://github.com/dbt-labs/dbt-cloud/pull/18139) but retracted.

dbtagarovat:
## Summary
- `Delete` on `dbtcloud_notification_setting` surfaced *any* API error as a hard failure, including a 404 returned when the resource is already gone server-side (reported as happening "sometimes" even though the delete had actually succeeded).
- This provider already has an established, tested pattern for exactly this case — `postgres_credential`, `spark_credential`, and `databricks_credential` all treat a `resource-not-found` error on delete as success via `helper.HandleResourceNotFound`. `notification_setting` was the outlier not using it.
- Fix: use `helper.HandleResourceNotFound` in `notification_setting`'s `Delete`, consistent with the rest of the provider, so a 404 no longer fails `terraform apply`.

## Verification
Also confirmed live against a real dbt Cloud backend (personal devspace), not just unit tests:
- `terraform apply` created a notification setting.
- Deleted it out-of-band via the API to simulate the "delete actually succeeded but client saw failure" scenario from the bug report.
- `terraform destroy` reported "no objects need to be destroyed" instead of erroring — confirmed both `Read` and `Delete` already handle the 404 correctly (no gap here, unlike the related fix in #724 where `Read` was missing this).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/framework/objects/notification_setting/...` (acceptance tests skip without `TF_ACC`, package compiles and unit-level checks pass)

Closes #718

### DIFF
--- a/.changes/unreleased/Fixes-20260803-132948.yaml
+++ b/.changes/unreleased/Fixes-20260803-132948.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: 'Make `dbtcloud_notification_setting` deletion idempotent: a 404 from the delete API is now treated as success instead of failing the apply (#718).'
+time: 2026-08-03T13:29:48.533476+03:00
--- a/pkg/framework/objects/notification_setting/behavior_test.go
+++ b/pkg/framework/objects/notification_setting/behavior_test.go
@@ -1,0 +1,69 @@
+package notification_setting
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/dbt-labs/terraform-provider-dbtcloud/pkg/dbt_cloud"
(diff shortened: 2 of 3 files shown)


tudor-scurtu (approved):
Thank you! :shipit:

PR #723, opened 2026-08-03
ci: retry transient dbt Cloud API failures in acceptance tests

dbtagarovat:
## Summary

Opened as a precursor to #720, #721, #722 — those PRs' `test-acceptance` check is currently failing for reasons unrelated to their changes, and this is why.

Investigated the CI acceptance-test failures (red on every daily run for 39 days, and on recent PRs). Ruled out:
- The retry-logic bug from #695 — already fixed by #696 a month before this started.
- A hard outage — `cloud.getdbt.com` is reachable and healthy, and **95%+ of individual acceptance tests pass in every single "failing" run** (342/353 on a recent PR run, 462/473 on 2026-06-26). This is intermittent flakiness at a measured ~2-3% per-test rate, not a dead backend.

Two concrete, in-repo fixes for that ~2-3%:

1. **`acctest_helper.SharedClient()` hardcoded `DisableRetry: true`.** This client is used by ~45 resource acceptance test files for post-apply "does this resource actually exist" / "was it actually destroyed" checks — unlike the real provider-under-test client (which retries 3x with backoff), every one of these verification calls was single-shot with zero tolerance for a transient blip. Found a live example in the CI logs: a bare `read: connection reset by peer` with zero retry attempts logged, failing a test whose actual resource management likely worked fine. Fixed to use the same retry policy (3 attempts, 10s backoff) as the main client. Added a regression test (`TestSharedClient_RetriesTransientErrors`) verified to fail against the old code and pass with the fix.
2. **Wrapped `make test-acceptance` with `nick-fields/retry`** (pinned to the exact commit for the `v3` tag, resolved via the GitHub API) in both `unit.yml` and `daily.yml`, retrying the whole suite once on failure. At a ~2-3% independent per-test failure rate, a genuine code regression still reproduces deterministically on retry and fails the build; a one-off transient timeout/502 does not.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/framework/acctest_helper/...` — new regression test passes with the fix, confirmed to fail without it
- [x] Validated both workflow YAML files parse correctly
- [ ] Confirm `test-acceptance` is green (or at least meaningfully less flaky) on this PR and on #720/#721/#722 after this merges

Marked as draft since I'd like to see this actually run against the live account before merging.

### DIFF
--- a/.changes/unreleased/Behind the scenes-20260803-160340.yaml
+++ b/.changes/unreleased/Behind the scenes-20260803-160340.yaml
@@ -1,0 +1,3 @@
+kind: Behind the scenes
+body: 'Reduce CI flakiness: retry the acceptance-test suite once on transient dbt Cloud API errors (~2-3% of tests hit single timeouts/502s per run against the shared test account), and fix `acctest_helper.SharedClient()` to use the same retry policy as the main provider client instead of `DisableRetry: true`, which made every acceptance test''s exists/destroy verification call single-shot with zero tolerance for a transient blip.'
+time: 2026-08-03T16:03:40.069232+03:00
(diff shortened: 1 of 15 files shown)


dbtagarovat:
Atm, mostly waiting and iterating on CI/CD, since draft didnt run through em. Willl give signal when ready proper

PR #724, opened 2026-08-03
fix: treat 404 on environment delete as idempotent success

dbtagarovat:
## Summary
- `environmentResource.Delete` returned a hard error for any API failure, including a 404 when the environment was already gone server-side (e.g. deleted via cascade from a related resource). That fails `terraform apply`/`destroy` even though the resource is, in effect, already gone.
- Same class of bug already fixed for `notification_setting` (#722): use `helper.HandleResourceNotFound` to treat a 404 as an idempotent success.
- Also fixed the same gap in `Read`: `Read` never called `HandleResourceNotFound` at all, so any 404 there was a hard error too. Since `terraform destroy`/`plan` always refresh state via `Read` before `Delete` runs, this broke destroy for a drifted-away environment even with the `Delete` fix in place (e.g. a CI/CD teardown retry after a prior timed-out request had actually already succeeded server-side).

## How this was found
While investigating acceptance-test flakiness for #723, `TestAccDbtCloudSalesforceCredentialDataSource`'s teardown failed deleting an environment. The logs showed the delete request actually got a real `404` response on its final retry attempt (not a timeout) - the environment had already been removed - but `Delete()` treated that 404 as a hard failure instead of success.

## Verification
Unit tests aside, this was also exercised live against a real dbt Cloud backend (personal devspace) to confirm both the bug and the fix, not just reasoned about from code:
- `terraform apply` created a project + environment; confirmed via direct API call.
- Deleted the environment out-of-band via the API (simulating the CI retry-after-timeout scenario from #723/#724's flakiness report).
- Confirmed `terraform destroy` failed at the refresh (`Read`) step before this fix, even though `Delete` was already idempotent — proving the `Read` gap was real and not just theoretical.
- After patching `Read` the same way, rebuilt and reran: `destroy` succeeded cleanly, and a follow-up API call confirmed the project was actually gone server-side.
- Regression-checked the normal path too: created fresh resources with no drift, ran `apply`/`destroy` end-to-end with no errors.

## Test plan
- [x] Added `TestDelete_404IsTreatedAsSuccess`, confirmed it fails on the pre-fix code and passes on the fix
- [x] Added `TestRead_404IsTreatedAsSuccess`, confirmed it fails on the pre-fix code and passes on the fix
- [x] `go build ./...`

### DIFF
--- a/.changes/unreleased/Fixes-20260803-175841.yaml
+++ b/.changes/unreleased/Fixes-20260803-175841.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: environment resource's Delete now treats a 404 (already deleted) as success instead of failing terraform apply
+time: 2026-08-03T17:58:41.056299+03:00
--- a/.changes/unreleased/Fixes-20260804-154217.yaml
+++ b/.changes/unreleased/Fixes-20260804-154217.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: environment resource's Read now treats a 404 (already deleted) as success instead of failing terraform plan/destroy, matching the existing Delete behavior
+time: 2026-08-04T15:42:17.000000+03:00
--- a/pkg/framework/objects/environment/behavior_test.go
+++ b/pkg/framework/objects/environment/behavior_test.go
@@ -1,0 +1,80 @@
+package environment
+
+import (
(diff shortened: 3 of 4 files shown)


victorasu (approved):
Nice!

PR #727, opened 2026-08-06
[Ad-Hoc] fix ado repo creation

victorasu:
Problem: Creating a `dbtcloud_repository` for ADO fails with 400.

Cause: The API doesn't accept ADO fields on update, only create. Even if the operation is Create, the provider also calls Update when `pull_request_url_template` is set. Error below.
`Additional properties are not allowed ('azure_active_directory_project_id', 'azure_active_directory_repository_id', 'azure_bypass_webhook_registration_failure' were unexpected)`

Solution: Since they are marked as RequireReplace anyway, we reset them to `nil` before marshaling, similar to the GitlabProjectID stopgap.

### DIFF
--- a/pkg/dbt_cloud/repository.go
+++ b/pkg/dbt_cloud/repository.go
@@ -190,6 +190,14 @@ func (c *Client) UpdateRepository(
 	// we need to remove the GitLab project ID for updates
 	repository.GitlabProjectID = nil
 
+	// The Azure DevOps fields are create-only (RequiresReplace in the schema) and
+	// are only accepted by the collection-create endpoint. The detail/update
+	// endpoint rejects them as unexpected properties ("Additional properties are
+	// not allowed"), so we strip them from update payloads.
+	repository.AzureActiveDirectoryProjectID = nil
+	repository.AzureActiveDirectoryRepositoryID = nil
+	repository.AzureBypassWebhookRegistrationFailure = nil
+
 	repositoryData, err := json.Marshal(repository)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


victorasu:
Failing tests are not related to change.

pkg/dbt_cloud/repository.go
```diff
@@ -190,6 +190,14 @@ func (c *Client) UpdateRepository(
 	// we need to remove the GitLab project ID for updates
 	repository.GitlabProjectID = nil
 
+	// The Azure DevOps fields are create-only (RequiresReplace in the schema) and
+	// are only accepted by the collection-create endpoint. The detail/update
+	// endpoint rejects them as unexpected properties ("Additional properties are
+	// not allowed"), so we strip them from update payloads.
+	repository.AzureActiveDirectoryProjectID = nil
+	repository.AzureActiveDirectoryRepositoryID = nil
+	repository.AzureBypassWebhookRegistrationFailure = nil
```
tudor-scurtu, line 199: My understanding is that this swallows the values silently.
shouldn't we keep the error for explicitness and have the user update their recipes to remove these values on non-create operations instead?

victorasu: It does swallow the values but after every operation we overwrite the state with the incoming values anyways. All we really do is just not provide them to the API, since they can't be changed in-place and need the current object to be removed and re-created.

It's not really good to have them do this, since resource definitions should stay the same unless the users want to change values, it's the provider's job to handle the subtleties of the API.

tudor-scurtu (approved):
Thank you! :shipit:

PR #734, opened 2026-08-23
fix: honor write-only semantic layer credential secrets

### LINKED ISSUE #732: dbtcloud_snowflake_semantic_layer_credential accepts private_key_wo but never sends it - credentials are created with an empty key, and rotating clears them
## Describe the bug
`dbtcloud_snowflake_semantic_layer_credential` accepts `private_key_wo` / `private_key_wo_version`, validates them, and its docs recommend them — but its Create and Update never read them. The credential is created or updated with an empty private key.

The resource reads only req.Plan, and only the plaintext fields ([snowflake_sl_credential_resource.go](https://github.com/dbt-labs/terraform-provider-dbtcloud/blob/main/pkg/framework/objects/semantic_layer_credential/snowflake_sl_credential_resource.go)):

```
resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
values := map[string]interface{}{
    "private_key":            plan.Credential.PrivateKey.ValueString(),
    "private_key_passphrase": plan.Credential.PrivateKeyPassphrase.ValueString(),
    // ...
}
```

There is no req.Config read and no reference to PrivateKeyWo anywhere in the file. Write-only values are null in the plan by design, so this returns "" whenever the configuration uses private_key_wo. The attributes exist only because the Snowflake SL schema and model reuse the standalone credential's (snowflake_credential.SnowflakeCredentialResourceSchema.Attributes) — which also inherits the validator that actively recommends switching to them.

Two distinct failures follow:

- On create, the apply succeeds and no later plan shows drift, because Read writes back only the IDs. It surfaces only when something queries, as JWT token is invalid from Snowflake.
- On update, the empty key is POSTed before the resource errors, so the API clears the stored key and the apply leaves the credential dead with state unrecorded. Bumping private_key_wo_version — the documented way to rotate — is the trigger. Because the POST replaces the credential record wholesale, it also drops the credential's service-token associations.

## Error message
On every plan, recommending the attribute that discards the key:
```
Warning: Available Write-Only Attribute Alternative

  with dbtcloud_snowflake_semantic_layer_credential.example,
  on credentials.tf line 89, in resource "dbtcloud_snowflake_semantic_layer_credential" "example":
  89:     private_key            = var.private_key

This attribute has a WriteOnly version credential.private_key_wo available.
Use the WriteOnly version of the attribute when possible.
```
On update, after the destructive write has already landed:
```

  .credential: inconsistent values for sensitive attribute

This is a bug in the provider.
```

That comes from Update setting state from the API response, which never echoes secrets back — the detail endpoint returns only auth_type, role, user, warehouse. The assertions are also unguarded, and will panic if those keys are absent rather than empty:

```
state.Credential.PrivateKey = types.StringValue(credential.Values["private_key"].(string))
```

## Resource configuration
```
resource "dbtcloud_snowflake_semantic_layer_credential" "example" {
  configuration = {
    project_id      = var.project_id
    name            = "SL - example"
    adapter_version = "snowflake_v0"
  }

  credential = {
    project_id                = var.project_id
    auth_type                 = "keypair"
    warehouse                 = "MY_WH"
    role                      = "MY_ROLE"
    user                      = "MY_SVC_USER"
    semantic_layer_credential = true

    private_key_wo         = var.private_key   # never sent
    private_key_wo_version = 1
  }
}
```
Applies cleanly. The credential cannot authenticate. Swapping in the plaintext private_key — the attribute the provider warns against — is the only configuration that works.

## Expected behavior
Either honor write-only attributes the way dbtcloud_snowflake_credential already does, or remove them from these resources so an unsupported configuration is rejected instead of silently discarded. The pattern is in the same repository, `pkg/framework/objects/snowflake_credential/resource.go`:

```
diags = req.Config.Get(ctx, &config)
private_key := helper.ResolveWriteOnlyString(config.PrivateKeyWo, plan.PrivateKey)
```

Update should also set state from the resolved configuration rather than the API echo, which resolves the inconsistent-result error and the unguarded assertions at the same time.

- dbt Cloud provider 1.12.3 and 1.12.5 — identical behavior; 1.12.5's copy of this file is byte-identical to main
- Terraform 1.15, auth_type = "keypair"
- snowflake_sl_credential_resource.go unchanged since #652 (3 Apr 2026), which predates write-only support, so no released version behaves differently

## Additional context
Four of the five SL credential resources in this repo look affected. BigQuery is also the only one that declares its *_wo attributes directly instead of inheriting them from a reused schema.

Production impact: one account, 20 credentials under a single for_each. Created with private_key_wo; all 20 held empty keys and nothing revealed it for nine days, because the apply succeeded and no plan showed drift. The keys were eventually installed by hand through the Admin API. Twelve days later, an apply bumping private_key_wo_version 1 → 2 cleared all 20 at once, along with every token mapping and errored — a 23-minute authentication outage, ended only by repairing every credential by hand again. 

### Workaround

Use the plaintext private_key and accept that it persists in state, then make every change a replacement via ignore_changes + replace_triggered_by — Create and Delete are the only operations this resource performs correctly.

pttydou:
Fixes #732.

The nested Snowflake, Redshift, Postgres, and Databricks semantic-layer credential resources read write-only secrets from the plan, where those values are unavailable. This change resolves write-only and regular inputs from configuration and plan, and keeps secret values out of update state.

Tests:
- `go test -p 1 -mod=readonly -count=1 ./...`
- `go build -p 1 -mod=readonly -ldflags "-w -s" .`
- `go vet ./pkg/framework/objects/semantic_layer_credential`
- `go generate ./...` and `git diff --exit-code -- docs`

Not run locally: `make test-acceptance` (requires private dbt Cloud and warehouse credentials).

### DIFF
--- a/.changes/unreleased/Fixes-20260823-225816.yaml
+++ b/.changes/unreleased/Fixes-20260823-225816.yaml
@@ -1,0 +1,3 @@
+kind: Fixes
+body: 'Send write-only secrets when creating or updating semantic layer credentials'
+time: 2026-08-23T22:58:16.1508412+08:00
--- a/pkg/framework/objects/semantic_layer_credential/databricks_sl_credential_resource.go
+++ b/pkg/framework/objects/semantic_layer_credential/databricks_sl_credential_resource.go
@@ -66,16 +66,22 @@ func (r *databricksSemanticLayerCredentialResource) Create(
 	req resource.CreateRequest,
 	resp *resource.CreateResponse,
 ) {
-	var plan DatabricksSLCredentialModel
+	var plan, config DatabricksSLCredentialModel
 
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	token := helper.ResolveWriteOnlyString(config.Credential.TokenWo, plan.Credential.Token)
 
 	values := map[string]interface{}{
 		"catalog": plan.Credential.Catalog.ValueString(),
-		"token":   plan.Credential.Token.ValueString(),
+		"token":   token,
 	}
 
 	createdCredential, err := r.client.CreateSemanticLayerCredential(
@@ -130,9 +136,13 @@ func (r *databricksSemanticLayerCredentialResource) Update(
 	req resource.UpdateRequest,
 	resp *resource.UpdateResponse,
 ) {
-	var plan, state DatabricksSLCredentialModel
+	var plan, state, config DatabricksSLCredentialModel
 
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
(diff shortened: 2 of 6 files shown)


stephan-king-ef:
Hello @pttydou , @VanTudor & @tudor-scurtu - just checking in on this PR. We have a workaround that lands the secrets in state, but we are very much looking forward to reverting that work and reintroducing the write-only approach.

zoltanersek:
@pttydou please rebase your changes onto main, we had a flaky build that we fixed and we need this branch to be up to date

PR #736, opened 2026-09-03
stabilize acceptance tests

zoltanersek:
- **retries with a body**: the retry loop reused a request whose body the first attempt had already consumed, so a retried create or update was sent empty and could never succeed
- **connection reuse**: the client used Go's default transport, which keeps 2 idle connections per host, so parallel calls re-dialed and re-negotiated TLS for most requests
- **check client**: `SharedClient` had retries disabled, so a single dropped response made `CheckDestroy` report a resource as still existing
- **leftovers**: a crashed run leaves an OpenAI integration or license map behind, and the account-level unique constraint then failed every later run; the owning tests clear them first
- **snowflake test**: its hardcoded OAuth client id collided with leftovers from earlier runs
- **runners**: acceptance tests move to Depot, and `go test -timeout 60m` so a stall fails a test instead of panicking before cleanup runs

Verified: `go build ./...`, `go vet ./pkg/...` and the unit suite pass. The retry and connection-reuse fixes each have a test that fails without them.


zoltanersek:
GH actions runner changes are not picked up by the PR build(because of the `pull_request_target` branch main setting) 
This was tested by running the workflow manually: https://github.com/dbt-labs/terraform-provider-dbtcloud/actions/runs/33760278486

zoltanersek:
closed in favor of https://github.com/dbt-labs/terraform-provider-dbtcloud/pull/737

PR #737, opened 2026-09-04
fix acceptance test flakiness

zoltanersek:
- run the test jobs on Depot, which removes the network failures the GitHub-hosted runners hit against the API
- randomize the Snowflake test's OAuth configuration client id, which has to be unique per account, so a leftover from an earlier run no longer fails every later one


zoltanersek:
GH actions runner changes are not picked up by the PR build(because of the pull_request_target branch main setting)
This was tested by running the workflow manually: https://github.com/dbt-labs/terraform-provider-dbtcloud/actions/runs/33858380380

tudor-scurtu (approved):
Thank you! :shipit:

PR #744, opened 2026-09-11
[CC-4349] Fix bigquery_credential Update() dropping WIF field changes

joaodaher:
## What's broken

`dbtcloud_bigquery_credential`'s `Update()` only ever sent `dataset`/`num_threads` changes to the dbt Cloud API. A change to `auth_type`, `workload_pool_provider_path`, or `service_account_impersonation_url` never reached the API — but `Update()` still wrote the plan into Terraform state as if it had, so state silently drifted from what the API actually had.

## The fix

Detect changes to the WIF fields too, and when the credential is adapter-based (v1), rebuild `credential_details` from the full plan the same way `Create()` already does, so the update call and the resulting state actually agree. Also added a missing `return` after the update-error diagnostic — every other error branch in this function has one; without it, a failed update call would still fall through and write the plan into state, which is the exact "state lies about the API" problem this PR is fixing, just via a different path.

## Worth knowing

A separate PR (#735, merged ~a week before this one) already added `RequiresReplace` to these three attributes, so a direct edit to a WIF field's *value* now goes through delete+recreate, not `Update()` — that specific failure mode is already closed. Where this fix still matters: `Update()` is the live path whenever `dataset`/`num_threads` change while the WIF fields stay the same, and previously that path would only round-trip `dataset`/`threads` through a GET-mutate-POST cycle without explicitly re-asserting the WIF fields. This closes that gap and matches what the ticket asked for.

## Test plan

Added a unit test asserting an update that changes the WIF fields actually reaches `UpdateBigQueryCredential` with those values (verified it fails against the pre-fix code). `go build`, `go vet`, and the package's unit tests are all clean; acceptance tests correctly skip without live credentials.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gravelbit (approved):
👌

PR #746, opened 2026-09-15
add api_endpoint to bigquery global connections

zoltanersek:
BigQuery connections could not use Private Link through Terraform. The provider accepted `private_link_endpoint_id`, but the connection still used the public BigQuery API, and the UI showed `Default endpoint`.

`api_endpoint` is the BigQuery API endpoint of the connection. Set it to the hostname of a Private Service Connect endpoint to send the traffic to that endpoint. `private_link_endpoint_id` records which endpoint the connection uses, so both fields are necessary. The provider did not have `api_endpoint`, although the API accepts it for `bigquery_v0` and for `bigquery_v1`. The provider already has the equivalent field for the other warehouses, such as `host`, `hostname`, `server` and `account`.

This PR adds the field to the resource and to the data source. A new example shows how to get both values from the `dbtcloud_privatelink_endpoint` data source.

The PR also has 2 documentation changes:

- `application_id` and `application_secret` are the client ID and the client secret for Native OAuth in development environments. The old text said that they applied only to `external-oauth-wif`, and the only example used WIF. A user can conclude that the fields do not exist. There is a new Native OAuth example.
- `private_link_endpoint_url` on the `dbtcloud_privatelink_endpoint` data source now says that the data source always returns it, so a lookup by name alone can supply the hostname of a connection. The attribute is also marked `Computed`, which matches `type` on the same data source and makes it plan as unknown if the read is deferred to apply. This does not change what a name-only lookup returns today: `TestPrivatelinkEndpointDataSource_PaginationTwoPages` already covered that case and passes either way.

I tested a local build against a dev account. I added `api_endpoint` to an existing connection, changed it, then removed it. I confirmed each step with a read of the connection API.


victorasu (approved):
Looks benign to me. :shipit:

docs/data-sources/privatelink_endpoint.md
```diff
@@ -37,7 +37,7 @@ data "dbtcloud_privatelink_endpoint" "test_with_name_and_url" {
 ### Optional
 
 - `name` (String) Given descriptive name for the PrivateLink Endpoint (name and/or private_link_endpoint_url need to be provided to return data for the datasource)
-- `private_link_endpoint_url` (String) URL of the PrivateLink Endpoint (name and/or private_link_endpoint_url need to be provided to return data for the datasource)
+- `private_link_endpoint_url` (String) URL of the PrivateLink Endpoint (name and/or private_link_endpoint_url need to be provided to return data for the datasource). Computed when only the name is provided, so that it can be used as the hostname of a connection.
```
victorasu: You mention here that it's computed in some cases, but the current changes do not cover this behavior. Was this a correction to the docs for existing logic?

zoltanersek: updated docs to make it clearer, the pr sets this as computed in the schema so it describes new changes

PR #747, opened 2026-09-17
fix the perpetual diff on environments with an empty deployment_type

zoltanersek:
An environment that sets `deployment_type = ""` never reaches a clean plan. Every plan after the apply asks for the same update:

```
actions: ["update"]
before.deployment_type: null
after.deployment_type : ""
```

A generic environment has no deployment type, and the API returns null for it. `Read` replaced the empty string from the config with that null, so the next plan tried to set it back. The apply then changed nothing: the update compares the two values as strings, and null and `""` are both `""`, so no request was sent and the state was rewritten from the plan. The next read flipped it to null again.

null and `""` both mean no deployment type, so `Read` now keeps `""` as it is. A value that really disappears upstream still shows as drift.

I tested a local build against a dev account. All four transitions converge, and a read of the environment API agrees at each step:

| transition | API after | next plan |
|---|---|---|
| `""` on create | null | No changes |
| `""` to `staging` | `staging` | No changes |
| `staging` to `""` | null | No changes |
| attribute removed | null | No changes |

`TestAccDbtCloudEnvironmentResourceEmptyDeploymentType` covers the same four transitions. It fails without the fix.


pkg/framework/objects/environment/resource.go
```diff
@@ -83,7 +83,15 @@ func (r *environmentResource) Read(
...
 		state.CustomBranch = types.StringPointerValue(environment.Custom_Branch)
 	}
-	state.DeploymentType = types.StringPointerValue(environment.DeploymentType)
+	// A generic environment has no deployment type, and the API returns null for it.
+	// An empty string in state means the same thing, so it is kept as it is. Otherwise a
+	// config that sets deployment_type = "" drifts to null on every read, and the plan
+	// that follows never converges.
+	if environment.DeploymentType != nil {
+		state.DeploymentType = types.StringValue(*environment.DeploymentType)
+	} else if state.DeploymentType.ValueString() != "" {
```
tudor-scurtu, line 92: Maybe I completely misunderstood, but shouldn't this be "=="?

tudor-scurtu (approved):
Thank you! approval to now slow you down, but left a question.

pkg/framework/objects/environment/resource.go
zoltanersek: That branch is only reached when the API returned no deployment type. The subtlety is that ValueString() collapses null and "" into the same "", so the condition means "state holds a real value"

victorasu (approved):

PR #749, opened 2026-09-21
add data sources for the account add-ons

zoltanersek:
dbt Wizard and dbt State are usage-based add-ons that an account turns on separately from its plan. The provider could not read them, and customers have asked for it.

## What this adds

- `dbtcloud_account_add_on` — the state of one product
- `dbtcloud_account_add_ons` — the state of every product the account can use

Both report the lifecycle state, the trial dates, whether the one trial per product is used up, and the spend limit that is set.

## Read only, on purpose

An earlier version of this PR also had a resource that turned products on, cancelled them and set the spend limit. It is gone.

Too much of the behaviour is not visible from the API, and none of it is visible from the configuration: paid activation can move the account onto a different plan, an allowlist decides whether it is allowed at all, and a spend limit cannot be set below what the current period already spent. A resource pinned to those rules would fail in ways a plan diff cannot explain, and the rules are still moving.

So the provider only reads. The spend limit is reported, and its description says to change it in the dbt platform. The tests register only GET handlers, so any write attempt fails the suite.

## Worth knowing

The two products treat the spend limit differently, and the docs say so: for `wizard` the limit stops usage at the limit, and for `state` nothing is stopped and it only raises a usage alert.

A product the account cannot use is left out of the API response entirely. That is not the same as a product the account has never turned on, which is still listed with no state. The single data source raises a clear error for the first case, and reports an empty state for the second.

## Tests

`data_source_acceptance_test.go` runs both data sources against a fake of the endpoints: every lifecycle state for both products, a product never turned on, a spend limit set, a limit of `0`, no limit, a product the account cannot use, and a list that carries only the available products.

## Checked against a dev account

| case | result |
|---|---|
| read `none`, `trial`, `active`, `cancelled`, `expired`, both products | all 10 map correctly through both data sources |
| a spend limit set outside Terraform | read back as `750000000000` |
| a product with no limit | reported as empty, not `0` |


victorasu (commented):
There's a few things I'm concerned with here.
My biggest is having a resource for add-ons. I think we should have a Data-Source for clients to have more visibility into their add-ons if they want it, but this object's attributes can change outside of the Terraform lifecycle for many reasons, which leads to a high potential of drift.

Even if we do decide to have a resource, a lot of the systems relying on the Add-on object have had flows designed around a certain order of operation (for example self-serve, after trial, need to set a spend limit to activate their add-on). Users should only be able to do what's possible in the UI, and in that order, nothing more. If we do decide to go ahead with a resource, we will need to look into exactly how each of the fields is set in the platform and what effects it has.

I'd honestly introduce the data source now, and introduce the resource if really necessary and only after carefully examining the potential side effects.

zoltanersek:
> There's a few things I'm concerned with here. My biggest is having a resource for add-ons. I think we should have a Data-Source for clients to have more visibility into their add-ons if they want it, but this object's attributes can change outside of the Terraform lifecycle for many reasons, which leads to a high potential of drift.
> 
> Even if we do decide to have a resource, a lot of the systems relying on the Add-on object have had flows designed around a certain order of operation (for example self-serve, after trial, need to set a spend limit to activate their add-on). Users should only be able to do what's possible in the UI, and in that order, nothing more. If we do decide to go ahead with a resource, we will need to look into exactly how each of the fields is set in the platform and what effects it has.
> 
> I'd honestly introduce the data source now, and introduce the resource if really necessary and only after carefully examining the potential side effects.

agreed, remove the resources for now

victorasu (approved):
The datasource alone looks benign enough 👌  :shipit:

PR #753, opened 2026-09-23
docs: Rename the fusion_admin permission label to v2 Migration Admin

andreiserbanzaha-dbt:
The `fusion_admin` permission set was renamed to **v2 Migration Admin** in the dbt platform UI, but our registry docs still showed "Fusion Admin", which [confused a client trying to match the UI to their Terraform config](https://dbt-labs.slack.com/archives/C08H1T53Y2F/p1789683525605959).

dbt-cloud renamed only the label (`V2_MIGRATION_ADMIN = "fusion_admin"`), so the value stays `fusion_admin`. Docs-only change: updated the permission table row in the `group` and `service_token` templates, regenerated `docs/` with `make doc`. No schema or validator change, no plan diff for existing configs.


victorasu (approved):
Lookin' fine!

